PR #29, opened 2017-04-25
Copy make test output to a file

johnbieren:
The build description is set to change based on the make test return code, but since groovy isn't in the environment yet, there is nothing to show.  Also, probably best to have this as its own file anyway.

### DIFF
--- a/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
+++ b/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
@@ -30,7 +30,7 @@ pushd $DIR_TO_GO
 # Run configure if it exists, if not, no big deal
 ./configure
 # Run tests if they are there
-make test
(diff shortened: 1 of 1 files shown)


config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -30,7 +30,7 @@ pushd $DIR_TO_GO
 # Run configure if it exists, if not, no big deal
 ./configure
 # Run tests if they are there
-make test
+make test >> ${OUTPUTDIR}/logs/make_test_output.txt
```
p3ck: will *.txt get archived?

johnbieren: The entire logs directory is rsynced and then logs/** is archived

PR #36, opened 2017-04-27
Added ostree image compose job

p3ck:
Should run approximately every 24 hours.

### DIFF
--- a/tasks/ostree-image-compose
+++ b/tasks/ostree-image-compose
@@ -1,0 +1,86 @@
+#!/bin/sh
+
+# Find the base dir
+base_dir="$(dirname $0)/.."
+
+for v in ostree images; do
+    rsync --delete --stats -a fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${BUILD}/${v}/ ${v}/
+done
+
+currentdir=$(pwd)
+
+# Clean up the imagefactory storage directory
+imgdir=/var/lib/imagefactory/storage/
+sudo rm -f $imgdir/*
+
+# Create directory for temporary files
+tmpdir=$(mktemp -d)
+
+# Kill backgrounded jobs on exit
+function clean_up {
+    kill $(jobs -p)
+    sudo rm -f $imgdir/*
+    sudo rm -rf $tmpdir
+}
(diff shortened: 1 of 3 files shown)


tasks/ostree-image-compose
```diff
@@ -0,0 +1,86 @@
...
+
+# Kill backgrounded jobs on exit
+function clean_up {
+    kill $(jobs -p)
+    sudo rm -f $imgdir/*
+    sudo rm -rf $tmpdir
+}
+trap clean_up EXIT SIGHUP SIGINT SIGTERM
+
+if [ ! -e "Fedora-Everything-netinst-x86_64-26-20170404.n.0.iso" ]; then
```
johnbieren: Is having this dependent on a date okay?

tasks/ostree-image-compose
```diff
@@ -0,0 +1,86 @@
...
+# Create a tdl file for imagefactory
+#       <install type='url'>
+#           <url>http://download.fedoraproject.org/pub/fedora/linux/releases/25/Everything/x86_64/os/</url>
+#       </install>
+cat <<EOF >./fedora-rawhide.tdl
+<template>
+    <name>rawhide</name>
+    <os>
+        <name>Fedora</name>
+        <version>22</version>
```
johnbieren: version 22? does this matter for anything?

tasks/ostree-image-compose
```diff
@@ -0,0 +1,86 @@
...
+        <arch>x86_64</arch>
+        <install type='iso'>
+            <iso>file://${currentdir}/Fedora-Everything-netinst-x86_64-26-20170404.n.0.iso</iso>
```
johnbieren: date here again, if it matters

tasks/ostree-image-compose
p3ck: These have been archived on the artifact server in centos so it should be ok.

tasks/ostree-image-compose
p3ck: nobody seems to update this in all the examples I've seen

tasks/ostree-image-compose
p3ck: we match what we downloaded above.  That's all that matters.

PR #53, opened 2017-05-04
Rpmbuild trigger fix

johnbieren:
So, currently we curl a url to find out if a branch is a development branch.  Turns out, that page has no fXX, just XX, so we need to strip the f off of the branch before checking.  I also added a job wide macro to determine a target release, so we will not ignore dist-git commits to that branch as well.

Note: Not a git pro, used multiple branches to make multiple PRs, this PR has a commit that removes a file a previous PR of mine added, but my intent is not to remove that file, I just didn't want that file to be part of this PR, so crossing my fingers that this PR does not remove that file if/when merged.


johnbieren:
Not sure if we would have to throttle this or not, I do not know the volume of commits to the now three branches per day.

PR #70, opened 2017-05-16
Add ostree compose boot sanity job.

p3ck:
This will boot the last generated qcow2 image and optionally
update to the passed in $COMMIT.  Actually, it doesn't do the update
yet, but it will soon. :-)

### DIFF
--- a/config/libvirt-setup/roles/libvirt-system/tasks/vm-import.yml
+++ b/config/libvirt-setup/roles/libvirt-system/tasks/vm-import.yml
@@ -1,0 +1,101 @@
+- name: VM template folder
+  action: file dest=/tmp/vm-{{ item.key }} state=directory
+
+- name: Figure out Base Image Name
+  set_fact:
+     base_image: "{{ item.value.img_url | basename | regex_replace('\\.(gz|xz)$', '') }}"
+
+- name: Get {{ item.value.img_url }}
+  get_url: url="{{ item.value.img_url }}" dest="/tmp/{{ item.value.img_url | basename }}"
+
+- stat: path="/tmp/{{ base_image }}.gz"
+  register: base_image_gz
+
+- stat: path="/tmp/{{ base_image }}.xz"
+  register: base_image_xz
+
+- name: Extract {{ base_image }}.xz"
+  command: xz -d /tmp/{{ base_image }}.xz
+  when: base_image_xz.stat.exists == True
+
+- name: Extract {{ base_image }}.gz"
+  command: gzip -d /tmp/{{ base_image }}.gz
+  when: base_image_gz.stat.exists == True
+
+- name: Install {{ base_image }} to LVM
+  command: qemu-img convert -O raw /tmp/{{ base_image }} /dev/{{ storage }}/lv_{{ item.key }}_root
+  when: storage is defined
+
+- name: Install {{ base_image }} to FILE
+  command: creates="{{ qemu_img_path }}/{{ item.key }}.img"
+           cp "/tmp/{{ base_image }}" "{{ qemu_img_path }}/{{ item.key }}.img"
+  when: qemu_img_path is defined
+
+- name: Remove {{ base_image }}
+  file: path="/tmp/{{ base_image }}" state=absent
+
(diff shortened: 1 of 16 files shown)


jobs/ostree-compose.yml
```diff
@@ -71,3 +71,31 @@
...
+        - ci-pipeline-duffy-builder:
+            task: ostree-boot-image
+            variables: |
+                export BUILD="${fed_branch}"
+                export image2boot="${image2boot}"
+                export commit="${commit}"
+                export JENKINS_JOB_NAME="${JOB_NAME}"
+                export JENKINS_BUILD_TAG="${BUILD_TAG}"
+                export OSTREE_BRANCH="${OSTREE_BRANCH:-}"
+            playbook: ci-pipeline/config/libvirt-setup/setup-libvirt.yml
```
johnbieren: This should be setup-system.yml, I don't see a setup-libvirt.yml

jobs/ostree-compose.yml
```diff
@@ -71,3 +71,31 @@
...
+              url pointing to the image to boot
+        - string:
+            name: commit
+            description: |
+              sha pointing to the particular commit we want to test
+    builders:
+        - ci-pipeline-duffy-builder:
+            task: ostree-boot-image
+            variables: |
+                export BUILD="${fed_branch}"
```
johnbieren: This is going to give you problems in ostree-boot-image because in that file you do not have a check to convert master to rawhide and you use it in the IMG_URL

config/libvirt-setup/setup-system.yml
```diff
@@ -0,0 +1,51 @@
+---
```
johnbieren: I'm not a pro on ansible roles, but can't libvirt-setup just be another role under config/duffy-setup?  Side note, I believe the plan when we move to jenkins pipeline is to provision one duffy node for the entire pipeline, so we are going to have to combine our requirements for the duffy host instead of having separate files.

config/libvirt-setup/roles/libvirt-system/tasks/vm-import.yml
```diff
@@ -0,0 +1,101 @@
...
+  pause: seconds=20
+  when: item.value.virt_customize is defined
+
+- name: Register hosts ip address
+  shell: arp -na | grep $(virsh dumpxml {{ item.key }} | xmllint --xpath 'string(//mac/@address)' -) | awk '{ print $2}' | sed 's/[\(\)]//g'
+  register: ip_address
+
+- name: Record hosts ip address
+  template: src="templates/hosts" dest="{{ inventory_dir }}/libvirt-hosts"
+  delegate_to: 127.0.0.1
```
johnbieren: I *might* request we change this at some point to anything but this because I think some of the atomic-host-tests playbooks do stuff with this address

config/libvirt-setup/roles/libvirt-system/templates/cloud-config/user-data
```diff
@@ -0,0 +1,25 @@
...
+users:
+  - name: admin
+    gecos: Admin User
+    groups: wheel
+    lock-passwd: false
+    passwd: {{ item.value.admin_passwd }}
+    ssh-authorized-keys:
+      - ssh-rsa {{ item.value.admin_ssh_rsa }}
+
+timezone: US/Eastern
```
johnbieren: (forgot to make this part of my review) can we use the timezone that ci.centos.org uses? Just so the console log timestamps in jenkins line up with test output from the libvirt host

jobs/ostree-compose.yml
```diff
@@ -68,6 +80,9 @@
                 export JENKINS_BUILD_TAG="${BUILD_TAG}"
                 export OSTREE_BRANCH="${OSTREE_BRANCH:-}"
             playbook: sig-atomic-buildscripts/centos-ci/setup/setup-system.yml
+        - inject:
+            properties-file: ${WORKSPACE}/logs/ostree.props
+        - ci-pipeline-ostree-boot-trigger
```
johnbieren: Not sure why you boot trigger twice

PR #71, opened 2017-05-18
Define fields for the fedmsg for rpmbuild

johnbieren:
This PR takes care of the logic for the fields in the defined fedmsgs, there is nothing here for the actual sending of the message as this needs to be set up via UI for now.

### DIFF
--- a/jobs/rpmbuild.yml
+++ b/jobs/rpmbuild.yml
@@ -57,6 +57,20 @@
     builders:
 #        - system-groovy:
 #            command: build.description = build.getEnvironment().get('fed_repo')
+        - shell: |
+            # Write ci message fields to a file
+            echo "topic=org.centos.prod.ci.pipeline.package.running" >> ${WORKSPACE}/job.properties_running
+            echo "build_url=$BUILD_URL" >> ${WORKSPACE}/job.properties_running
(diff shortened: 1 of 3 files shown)


jobs/rpmbuild.yml
```diff
@@ -57,6 +57,20 @@
...
+            # Write ci message fields to a file
+            echo "topic=org.centos.prod.ci.pipeline.package.running" >> ${WORKSPACE}/job.properties_running
+            echo "build_url=$BUILD_URL" >> ${WORKSPACE}/job.properties_running
+            echo "build_id=$BUILD_ID" >> ${WORKSPACE}/job.properties_running
+            echo "ref=fedora/${fed_branch}/x86_64/atomic-host" >> ${WORKSPACE}/job.properties_running
+            echo "rev=$fed_rev" >> ${WORKSPACE}/job.properties_running
+            echo "namespace=rpms" >> ${WORKSPACE}/job.properties_running
+            echo "repo=$fed_repo" >> ${WORKSPACE}/job.properties_running
+            echo "status=success" >> ${WORKSPACE}/job.properties_running
+            echo "test_guidance=''" >> ${WORKSPACE}/job.properties_running
```
p3ck: this is fine but I think a cat <<EOF > ${WORKSPACE}/job.properties_running is cleaner.

PR #74, opened 2017-05-18
Add jobs to provide framework to run tests on pipeline built kernels

johnbieren:
All this does for now is check that the kernel installs, but eventually we will put more testing here (hopefully ltp/lite soon)

### DIFF
--- a/tasks/kernel-kt0
+++ b/tasks/kernel-kt0
@@ -1,0 +1,77 @@
+#!/bin/sh
+
+# Find the base dir
+base_dir="$(dirname $0)/.."
+
+mkdir -p logs
+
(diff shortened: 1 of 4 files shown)


tasks/kernel-kt0
```diff
@@ -40,11 +41,14 @@ EOF
 # Cannot be part of last EOF since IMG_URL needs to resolve
 cat << EOF >> host_vars/kernel_install_slave_baremetal.yml
    img_url: $IMG_URL
+   admin_ssh_rsa: $pubkey
```
p3ck: You forgot to remove it from the previous cat EOF section

PR #93, opened 2017-05-22
mkdir: cannot create directory ‘host_vars’: File exists

johnbieren:
Saw the error from subject in console output log, so thinking we need this

### DIFF
--- a/tasks/kernel-kt0
+++ b/tasks/kernel-kt0
@@ -27,8 +27,7 @@ fi
 
 pubkey=$(cat ~/.ssh/id_rsa.pub)
 
-rm -rf host_vars
-mkdir host_vars
+mkdir -p host_vars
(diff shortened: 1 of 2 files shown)


tasks/kernel-kt0
```diff
@@ -27,6 +27,7 @@ fi
 
 pubkey=$(cat ~/.ssh/id_rsa.pub)
 
+rm -rf host_vars
 mkdir host_vars
```
p3ck: or mkdir -p host_vars

tasks/ostree-boot-image
```diff
@@ -48,6 +48,7 @@ fi
 
 pubkey=$(cat ~/.ssh/id_rsa.pub)
 
+rm -rf host_vars
 mkdir host_vars
```
p3ck: same

PR #103, opened 2017-05-23
Use per branch kickstarts and use a pacthed version we supply instead.

p3ck:


tasks/ostree-image-compose
```diff
@@ -63,7 +63,8 @@ popd
 pushd $tmpdir
 
 # Grab the kickstart file
-curl -O -L https://pagure.io/fedora-kickstarts/raw/master/f/fedora-atomic.ks
+#curl -O -L https://pagure.io/fedora-kickstarts/raw/master/f/fedora-atomic.ks
+cp $currentdir/ci-pipeline/config/ostree/fedora-atomic-${BRANCH}.ks fedora-atomic.ks
```
johnbieren: I am pretty sure you want ${BUILD} here, not ${BRANCH}

p3ck: yup - thanks!

PR #112, opened 2017-05-24
fix test_guidance again

johnbieren:
test_guidance is giving me grief again with the cat EOF method so I think adding the quotes here fixes it.


johnbieren:
Well, the first build that executed with this changed stopped getting null pointer exception after like six in a row, so I don't know maybe something else was broken at the same time.  I will try to break it again I guess

johnbieren:
It seems it does not reproduce the null pointer exception 100% of the time, but if you compare 
https://ci.centos.org/job/ci-pipeline-rpmbuild/701/
with
https://ci.centos.org/job/ci-pipeline-rpmbuild/702/
the *only* thing I changed was adding the "" as described in this PR

PR #113, opened 2017-05-24
Small fixes to kernel and atomic-host-tests tasks

johnbieren:


johnbieren:
No, the ~/.ssh/id_rsa does not do anything; you are right.  Still trying to find the fix

PR #115, opened 2017-05-24
Publish messages on the bus at runtime and completion.

p3ck:
messages aded for ostree-compose, ostree-compose-image and ostree-compose-boot-sanity

### DIFF
--- a/jobs/ostree-compose.yml
+++ b/jobs/ostree-compose.yml
@@ -1,42 +1,45 @@
-- builder:
+- publisher:
     name: ci-pipeline-ostree-boot-trigger
-    builders:
-        - trigger-builds:
+    publishers:
+        - trigger-parameterized-builds:
             - project: ci-pipeline-ostree-boot-sanity
               current-parameters: true
               property-file: ${WORKSPACE}/logs/ostree.props
-              block: true
+              block: false
               same-node: true
 
-- builder:
+- publisher:
     name: ci-pipeline-ostree-image-trigger
-    builders:
-        - conditional-step:
-            condition-kind: shell
-            condition-command: |
-                prev=0
-                number=$(curl $JENKINS_URL/job/{compose_job}/lastBuild/api/json 2>/dev/null | jq '.number')
-                for ((i=$number; i>0; i--)); do
-                    branch=$(curl $JENKINS_URL/job/{compose_job}/$i/api/json 2>/dev/null | jq -r '.actions[] | select(.parameters).parameters[] | select(.name | contains("fed_branch")).value')
-                    if [ "$branch" = "$fed_branch" ]; then
-                        result=$(curl $JENKINS_URL/job/{compose_job}/$i/api/json 2>/dev/null| jq -r '.result')
-                        if [ "$result" = "SUCCESS" ]; then
-                            prev=$(curl $JENKINS_URL/job/{compose_job}/$i/api/json 2>/dev/null| jq '.timestamp' | cut -c1-10)
-                            break
+    publishers:
+        - postbuildscript:
+            builders:
+                - conditional-step:
+                    condition-kind: shell
+                    condition-command: |
+                        prev=0
+                        number=$(curl $JENKINS_URL/job/{compose_job}/lastBuild/api/json 2>/dev/null | jq '.number')
+                        for ((i=$number; i>0; i--)); do
+                            branch=$(curl $JENKINS_URL/job/{compose_job}/$i/api/json 2>/dev/null | jq -r '.actions[] | select(.parameters).parameters[] | select(.name | contains("fed_branch")).value')
+                            if [ "$branch" = "$fed_branch" ]; then
+                                result=$(curl $JENKINS_URL/job/{compose_job}/$i/api/json 2>/dev/null| jq -r '.result')
+                                if [ "$result" = "SUCCESS" ]; then
+                                    prev=$(curl $JENKINS_URL/job/{compose_job}/$i/api/json 2>/dev/null| jq '.timestamp' | cut -c1-10)
+                                    break
+                                fi
+                            fi
+                        done
+                        cur=$(date +%s)
+                        elapsed=$((cur - prev))
+                        if [ $elapsed -gt 86400 ]; then
+                           exit 0
                         fi
-                    fi
-                done
-                cur=$(date +%s)
-                elapsed=$((cur - prev))
-                if [ $elapsed -gt 86400 ]; then
-                   exit 0
-                fi
-                exit 1
-            on-evaluation-failure: dont-run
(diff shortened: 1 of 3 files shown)


jobs/ostree-compose.yml
```diff
@@ -57,13 +104,43 @@
             playbook: sig-atomic-buildscripts/centos-ci/setup/setup-system.yml
         - inject:
             properties-file: ${WORKSPACE}/logs/ostree.props
+    publishers:
+        - ci-pipeline-duffy-publisher
+        - postbuildscript:
+            builders:
+                - shell: |
```
johnbieren: I don't get what the point of writing the variables to a file and injecting them is if there is no shell logic here, just variable assignment and the assigned values can just be used in the jms-messaging stanza like "status": "${BUILD_STATUS}" instead

jobs/ostree-compose.yml
```diff
@@ -83,10 +206,45 @@
             playbook: sig-atomic-buildscripts/centos-ci/setup/setup-system.yml
         - inject:
             properties-file: ${WORKSPACE}/logs/ostree.props
-        - ci-pipeline-ostree-boot-trigger
-
     publishers:
         - ci-pipeline-duffy-publisher
+        - postbuildscript:
+            builders:
+                - shell: |
```
johnbieren: see above

jobs/ostree-compose.yml
```diff
@@ -118,6 +322,39 @@
             playbook: sig-atomic-buildscripts/centos-ci/setup/setup-system.yml
     publishers:
         - ci-pipeline-duffy-publisher
+        - postbuildscript:
+            builders:
```
johnbieren: see above

jobs/ostree-compose.yml
```diff
@@ -59,40 +59,30 @@
...
-                test_guidance=
-            EOF
+            if [ "$fed_branch" = "master" ]; then
+                echo "fed_branch=rawhide" > ${WORKSPACE}/job.properties
+            fi
         - inject:
-            properties-file: ${WORKSPACE}/job.properties_running
+            properties-file: ${WORKSPACE}/job.properties
         - jms-messaging:
             override-topic: org.centos.prod.ci.pipeline.compose
```
johnbieren: So I think in the cases where you know topic (which here you do as per line 72), you can put the full topic here.  The reason why I abridged mine was because mine is variable (mainly because it could be .ignore .queued etc), and $topic is not supported in override-topic yet, so I had to leave off the end (that was my thinking at least, I could have left it cut off in unnecessary places in my code).

jobs/ostree-compose.yml
```diff
@@ -106,37 +96,26 @@
...
-                            "ref": "${ref}",
-                            "rev": "${rev}",
-                            "namespace": "${namespace}",
-                            "repo": "${repo}",
-                            "status": "${status}",
-                            "test_guidance": "${test_guidance}"
-                        }
-            script-only-if-succeeded: false
+        - jms-messaging:
+            override-topic: org.centos.prod.ci.pipeline.compose
```
johnbieren: above

jobs/ostree-compose.yml
```diff
@@ -155,46 +134,33 @@
...
-                type=qcow2
-            EOF
+            if [ "$fed_branch" = "master" ]; then
+                echo "fed_branch=rawhide" > ${WORKSPACE}/job.properties
+            fi
         - inject:
-            properties-file: ${WORKSPACE}/job.properties_running
+            properties-file: ${WORKSPACE}/job.properties
         - jms-messaging:
             override-topic: org.centos.prod.ci.pipeline.image
```
johnbieren: above

jobs/ostree-compose.yml
```diff
@@ -208,42 +174,29 @@
...
-                            "repo": "${repo}",
-                            "compose_url": "${compose_url}",
-                            "compose_rev": "${compose_rev}",
-                            "status": "${status}",
-                            "test_guidance": "${test_guidance}",
-                            "type": "${type}"
-                        }
-            script-only-if-succeeded: false
+        - jms-messaging:
+            override-topic: org.centos.prod.ci.pipeline.image
```
johnbieren: above

jobs/ostree-compose.yml
```diff
@@ -268,46 +221,33 @@
...
-                type=qcow2
-            EOF
+            if [ "$fed_branch" = "master" ]; then
+                echo "fed_branch=rawhide" > ${WORKSPACE}/job.properties
+            fi
         - inject:
-            properties-file: ${WORKSPACE}/job.properties_running
+            properties-file: ${WORKSPACE}/job.properties
         - jms-messaging:
             override-topic: org.centos.prod.ci.pipeline.image.test.smoke
```
johnbieren: above

jobs/ostree-compose.yml
```diff
@@ -322,39 +262,29 @@
...
-                            "repo": "${repo}",
-                            "compose_url": "${compose_url}",
-                            "compose_rev": "${compose_rev}",
-                            "status": "${status}",
-                            "test_guidance": "${test_guidance}",
-                            "type": "${type}"
-                        }
-            script-only-if-succeeded: false
+        - jms-messaging:
+            override-topic: org.centos.prod.ci.pipeline.image.test.smoke
```
johnbieren: above

jobs/ostree-compose.yml
p3ck: so can mine.  I don't understand any difference.  unless you have a better explanation I'm going to leave it as is.

thanks

johnbieren: Yours is not variable.  This is your code:
topic=org.centos.prod.ci.pipeline.compose.running
That is static.  It will never change. So in this publisher, the override-topic can match it.  Mine is not static.  This is mine
topic=${{topic}}
Which I get from injecting.  However, you cannot use ${{topic}} in override-topic just yet, so I cut off my topic in the override-topic line.  Not sure I follow you how yours can change when the topic doesn't have any variables in it, just static text

p3ck: why do you echo everything out to a properties file only to read it back in?  Isn't this exactly what you complained about in this PR of mine? Don't get me wrong, I agree with your assessment and that's why I changed it.

arilivigni:
LGTM

jobs/ostree-compose.yml
johnbieren: I do it in rpmbuild.yml because I use jms-messaging twice, and if I ever have to change anything I only want to have to change it in one place.  But you are 100% correct, it is entirely unnecessary in rpmbuild_trigger.yml and I will change it.

PR #138, opened 2017-05-31
Allow rpmbuild jobs to be concurrent

johnbieren:
We have 24 executors now in the CentOS ci environment, so I would like to see if concurrency works.  This would be a big help for rpmbuild, as kernel builds tend to take multiple hours and generate quite a queue.  Unsure if this will *just work* or will need further changes, so this is one way to find out.

### DIFF
--- a/jobs/rpmbuild.yml
+++ b/jobs/rpmbuild.yml
@@ -1,6 +1,7 @@
 - job:
     name: ci-pipeline-rpmbuild
     defaults: ci-pipeline-defaults
+    concurrent: true
     parameters:
         - string:
             name: fed_username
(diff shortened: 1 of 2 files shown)


jobs/rpmbuild.yml
```diff
@@ -1,6 +1,7 @@
 - job:
     name: ci-pipeline-rpmbuild
     defaults: ci-pipeline-defaults
+    concurrent: true
```
p3ck: what about the manifest file?

johnbieren: Interesting point, I didn't consider that.  I will have to devise a solution

PR #150, opened 2017-06-07
Initial commit of beakerlib ansible module

johnbieren:
So far, this ansible module seems to work for rlRun, rlAssertGrep, and rlAssertNotGrep from the command line execution of it as described in the 'Examples' section.  Some things to think about are what other beakerlib functions we want to add, the output we give, and whether or not we want to support running rlRun commands the shell way instead of the command way, as I believe this only does it the command way since I took from that module's code.  Regarding output, an example is I do not currently give as output the command you ran with rlRun, nor the file or string parameters you input.  Not sure how necessary that would be.  We also would likely eventually need some checks to make sure arguments are provided when they are required.

### DIFF
--- a/library/beakerlib.py
+++ b/library/beakerlib.py
@@ -134,12 +134,12 @@ def main():
         argument_spec=dict(
           _raw_params = dict(),
           _uses_shell = dict(type='bool', default=False),
-          command = dict(),
+          command = dict(default=None),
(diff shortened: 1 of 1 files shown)


johnbieren:
Looking a little more at various modules I believe I need to make some slight modifications to this so that the arguments do not need to be under args: and can just be normal. Update coming soon

PR #152, opened 2017-06-08
Jenkins 2.0 Pipeline file

arilivigni:
- Converted all individual jobs from JJB into pipeline
- Checking for variables
- Make libraries out of repeated functions

Note: This will probably need to be refactored as we get used to using pipelines
 - Remove echo and prints
 - reduce redundant defs of vars where possible

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,8 +23,6 @@ node('fedora-atomic') {
                 stage('ci-pipeline-rpmbuild-trigger') {
                     def current_stage = "ci-pipeline-rpmbuild-trigger"
 
-                    echo "Current Stage is: ${current_stage}"
-
                     // Python script to parse the ${CI_MESSAGE}
                     writeFile file: "${env.WORKSPACE}/parse_fedmsg.py",
                               text: "#!/bin/env python\n" +
@@ -63,10 +61,9 @@ node('fedora-atomic') {
                         branch=${fed_branch}
                         if [ "${branch}" = "master" ]; then
                           branch="rawhide"
-                          fed_branch="rawhide"
                         fi
                         echo "branch=${branch}" >> ${WORKSPACE}/job.properties
-                        echo "namespace=${namespace}" >> ${WORKSPACE}/job.properties
+                        #echo "namespace=${namespace}" >> ${WORKSPACE}/job.properties
                         
                         # Verify this is a branch in our list of targets defined above in the parameters
                         if [[ ! "${fed_branch}" =~ "${TARGETS}" ]]; then
@@ -95,7 +92,7 @@ node('fedora-atomic') {
                     convertProps(job_props, job_props_groovy)
                     load(job_props_groovy)
 
-                    // Send message on fedmsg
+                    // Send message org.centos.prod.ci.pipeline.package.queued or .ignore on fedmsg
                     sendCIMessage messageContent: '',
                             messageProperties: "topic=${topic}\n" +
                                     "build_url=${BUILD_URL}\n" +
@@ -104,7 +101,7 @@ node('fedora-atomic') {
                                     "ref=fedora/${branch}/x86_64/atomic-host\n" +
                                     "rev=${fed_rev}\n" +
                                     "repo=${fed_repo}\n" +
-                                    "namespace=${namespace}\n" +
+                                    "namespace=${fed_namespace}\n" +
                                     "username=fedora-atomic\n" +
                                     "test_guidance=''\n" +
                                     "status=success",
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,497 @@
...
+                    // Check if package is in the package list for fedora-atomic host
+                    sh '''
+                        set +e
+                        branch=${fed_branch}
+                        if [ "${branch}" = "master" ]; then
+                          branch="rawhide"
+                          fed_branch="rawhide"
+                        fi
+                        echo "branch=${branch}" >> ${WORKSPACE}/job.properties
+                        echo "namespace=${namespace}" >> ${WORKSPACE}/job.properties
```
johnbieren: You obviously know way more about what variables are defined and when in the jenkinsfile but is ${namespace} defined here? My thought is that it would be fed_namespace, but again I am likely wrong

Jenkinsfile
```diff
@@ -0,0 +1,497 @@
...
+                        
+                        # Verify this is a branch in our list of targets defined above in the parameters
+                        if [[ ! "${fed_branch}" =~ "${TARGETS}" ]]; then
+                            echo "${fed_branch} is not in the list"
+                            echo "topic=org.centos.prod.ci.pipeline.package.ignore" >> ${WORKSPACE}/job.properties
+                        elif
```
johnbieren: Does elif work with no check? Would this be else instead? It may very well work, I have just never seen elif with no boolean check

Jenkinsfile
```diff
@@ -0,0 +1,497 @@
...
+                        // Stage resources - RPM build system
+                        setupStage()
+
+                        // Rsync Data
+                        writeFile file: "${env.ORIGIN_WORKSPACE}/task.env",
+                                  text: "export JENKINS_JOB_NAME=\"${JOB_NAME}-${current_stage}\"\n" +
+                                        "export JENKINS_BUILD_TAG=\"${BUILD_TAG}-${current_stage}\"\n" +
+                                        "export OSTREE_BRANCH=\"\${OSTREE_BRANCH:-}\"\n" +
+                                        "export fed_repo=\"${fed_repo}\"\n" +
+                                        "export fed_branch=\"${fed_branch}\"\n" +
```
johnbieren: In line 66 we changed fed_branch to rawhide if it was master.  Is it still rawhide here?  Not sure about variables in the different stages.  If it is still master and the variable change persists, we need to not change fed_branch in line 66.  The reason for this is that the rpmbuild Dockerfile entry point relies on fed_branch being master for the rawhide case, as it does a git checkout ${fed_branch}

arilivigni: - I will check the namespace I saw many uses of fed_<var> and just <var> it was hard to keep track
- That should have been an else my mistake.
- When it comes to the branch if we set it at the top from master to rawhide my intent it would carry all the way through.  Is that not the case?

johnbieren: In the current workflow, if fed_branch = rawhide we create the new variable branch which points to master.  The reason why we create that branch variable instead of simply overwriting fed_branch is because for example the rpmbuild uses "master" instead of rawhide.  If we did it properly, for f27 right now, branch should be rawhide and fed_branch should be master throughout to the best of my knowledge

Jenkinsfile
```diff
@@ -0,0 +1,497 @@
...
+                        def ostree_props_groovy = "${env.ORIGIN_WORKSPACE}/ostree.props.groovy"
+                        convertProps(ostree_props, ostree_props_groovy)
+                        load(ostree_props_groovy)
+
+
+                        // Check if a new ostree image compose is needed
+                        checkLastImage()
+                        if (fileExists("${env.WORKSPACE}/NeedNewImage.txt")) {
+                            stage("ci-pipeline-ostree-image-compose") {
+                                current_stage = "ci-pipeline-ostree-image-compose"
```
p3ck: Is this just a place holder?  It doesn't look like image compose or boot sanity are implemented yet?

Jenkinsfile
arilivigni: Gotcha I will make those changes then to ensure we keep fed_branch the way it is.  It looks like Bill converted it in his jobs for ostree composes

Jenkinsfile
arilivigni: Yes it is more to be added shortly

arilivigni:
@p3ck @johnbieren updated to address your concerns.  I realize there is some commonality in the messages and other pieces that I will reduce but I rather you look through and check what is there.

Jenkinsfile
arilivigni: @p3ck this is updated in the latest patch.  I also did fill in rev and repo to use fed_rev and fed_repo in the message

PR #159, opened 2017-06-15
Changes to pipeline and service job

arilivigni:
- Removed extra defining of current_stage on teardown
- Remove quotes around ${TARGETS}
- change service_jobs to force update --ignore-cache

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -69,7 +69,7 @@ node('fedora-atomic') {
                         echo "branch=${branch}" >> ${WORKSPACE}/job.properties
                         
                         # Verify this is a branch in our list of targets defined above in the parameters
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -69,7 +69,7 @@ node('fedora-atomic') {
                         echo "branch=${branch}" >> ${WORKSPACE}/job.properties
                         
                         # Verify this is a branch in our list of targets defined above in the parameters
-                        if [[ ! "${fed_branch}" =~ "${TARGETS}" ]]; then
```
scoheb: why do we need to remove the quotes?

arilivigni: It fails with them.  I tested it even in a shell script

scoheb: lgtm then

PR #164, opened 2017-06-20
enable concurrency for rpmbuilding and atomic testing

johnbieren:

### DIFF
--- a/tasks/rpmbuild-test
+++ b/tasks/rpmbuild-test
@@ -25,6 +25,24 @@ if [ "${fed_branch}" = "master" ]; then
 fi
 mkdir ${RSYNC_BRANCH}
 mkdir repo
+# Write uuid to a lock file and store a backup
+uuidgen > file.lock
+cp file.lock uuid.saved
+while false; do
+     # Check if lock exists on remote server
+     while [[ $(rsync --dry-run -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir) != *"file.lock"* ]]; do
+          sleep 60
+     done
+     # Push lock file with uuid to remote server
+     rsync -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/
+     # Pull lock file back
+     rsync -avz fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/file.lock file.lock
+     # If uuid matches, we can proceed
+     if [[ $(diff file.lock uuid.saved) == "" ]]; then
+          break
+     fi
+     sleep 60
+done
 # Rsync the empty directories over first, then the repo directory
 rsync -arv ${RSYNC_BRANCH}/ fedora-atomic@artifacts.ci.centos.org::fedora-atomic
 rsync -arv repo/ fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}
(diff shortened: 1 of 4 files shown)


tasks/rpmbuild-test
```diff
@@ -25,6 +25,24 @@ if [ "${fed_branch}" = "master" ]; then
 fi
 mkdir ${RSYNC_BRANCH}
 mkdir repo
+# Write uuid to a lock file and store a backup
+uuidgen > file.lock
+cp file.lock uuid.saved
+while false; do
```
p3ck: This will never run...  Don't you want:
`while true; do`

tasks/rpmbuild-test
```diff
@@ -25,6 +25,24 @@ if [ "${fed_branch}" = "master" ]; then
...
+     # Check if lock exists on remote server
+     while [[ $(rsync --dry-run -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir) != *"file.lock"* ]]; do
+          sleep 60
+     done
+     # Push lock file with uuid to remote server
+     rsync -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/
```
p3ck: You should also add the following rsync args to prevent a race condition when transferring files to the remote
`--ignore-existing`

tasks/rpmbuild-test
```diff
@@ -39,5 +57,7 @@ sed -i "/${fed_repo}_repo/d" manifest.txt
 echo "${fed_repo}_repo $(date --utc +%FT%T%Z)" >> manifest.txt
 sort manifest.txt -o manifest.txt
 rsync --delete -stats -a manifest.txt fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo
+# Delete the rsync lock we placed
+rsync -vr --delete $(mktemp -d)/ fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/
```
p3ck: this should be done in the trap handler so that we always remove the lock even if we abort for some reason.

tasks/rpmbuild-test
```diff
@@ -25,6 +25,24 @@ if [ "${fed_branch}" = "master" ]; then
...
+     # Push lock file with uuid to remote server
+     rsync -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/
+     # Pull lock file back
+     rsync -avz fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/file.lock file.lock
+     # If uuid matches, we can proceed
+     if [[ $(diff file.lock uuid.saved) == "" ]]; then
+          break
+     fi
+     sleep 60
+done
```
p3ck: You should define a trap function to remove the lock after getting it.  Do this instead of lines 60 and 61.  This way if anything breaks the script it will still remove the lock file.
```
# Kill backgrounded jobs on exit
function clean_up {
# Delete the rsync lock we placed
rsync -vr --delete $(mktemp -d)/ fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/
}
trap clean_up EXIT SIGHUP SIGINT SIGTERM

```

tasks/rpmbuild-test
johnbieren: Yea, thanks for pointing that out

PR #170, opened 2017-06-21
Fix locks and add some sudo to docker commands

johnbieren:

### DIFF
--- a/tasks/rpmbuild-test
+++ b/tasks/rpmbuild-test
@@ -13,7 +13,8 @@ pushd $base_dir/config/Dockerfiles/rpmbuild
 which docker
 if [ "$?" != 0 ]; then echo "ERROR: DOCKER NOT INSTALLED\nSTATUS: $?"; exit 1; fi
 # Only build if it is not built already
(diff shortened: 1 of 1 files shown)


tasks/rpmbuild-test
```diff
@@ -30,25 +31,26 @@ uuidgen > file.lock
 cp file.lock uuid.saved
 while true; do
      # Check if lock exists on remote server
-     while [[ $(rsync --dry-run -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir) != *"file.lock"* ]]; do
+     while [[ $(rsync --ignore-existing --dry-run -avz file.lock fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir) != *"file.lock"* ]]; do
           sleep 60
      done
+     cp uuid.saved file.lock
```
p3ck: this is pretty hacky...  I think I would combine the above rsync with an -e "file.lock" test.

johnbieren: That isn't my point here.  If you push your file.lock out, then pull it back and it doesn't match because you got someone else's file.lock, you now do not have your own uuid in file.lock for the next iteration of the loop.  Or ever, for that matter

PR #174, opened 2017-06-22
default time zone for Jenkins instance

scoheb:


config/s2i/jenkins/master/configuration/init.groovy
```diff
@@ -11,9 +11,14 @@ logger.info("Enable Slave -> Master Access Control")
 Jenkins.instance.injector.getInstance(AdminWhitelistRule.class).setMasterKillSwitch(false);
 Jenkins.instance.save()
 
+logger.info("Setting Time Zone to be EST")
```
pypingou: Having this be UTC may also be an idea considering UTC never changes.

PR #181, opened 2017-06-30
Use containers for the heavy lifting

p3ck:
moved ostree compose to a container
moved rsync process to a container


config/Dockerfiles/ostree_compose/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN yum -y localinstall https://kojipkgs.fedoraproject.org//packages/python-distro/1.0.1/2.el7/noarch/python2-distro-1.0.1-2.el7.noarch.rpm
+RUN yum -y install --disablerepo=epel-testing rsync mock libsolv glib2 ostree rpm-ostree rpm-ostree-toolbox fedpkg PyYAML rpmdistro-gitoverlay libgsystem genisoimage ansible
+
+RUN git clone https://github.com/CentOS-PaaS-SIG/ci-pipeline
+
+# Copy the build script to the container
+COPY ostree-compose.sh $HOME/ci-pipeline/
+
+# Run the build script
+ENTRYPOINT ["bash", "/root/ci-pipeline/ostree-compose.sh"]
```
johnbieren: Since you have ENV HOME=/root, do you want to use $HOME/ci-pipeline/ostree-compose.sh to stay consistent?  Same comment for the rsync container.

p3ck: I tried that but it didn't resolve $HOME.  Maybe I should have done it like this??
`ENTRYPOINT ["bash", $HOME"/ci-pipeline/ostree-compose.sh"]`

johnbieren: Even if that works I think that is kind of ugly.  I'd just leave it as is then, didn't know you had tried that.

PR #185, opened 2017-07-07
Add docker container to run package level testing against an ansible host

johnbieren:
This container takes in an inventory file and a private key as inputs and runs package level testing against the ansible host.  It is set up to require both the ssh private key and the inventory file to be in the same directory on the host it is called from.  It also writes the artifacts out to that same directory.

Also fixes a small check in the rpmbuild task for deciding whether or not to build the image that Bill pointed out.

### DIFF
--- a/config/Dockerfiles/package_tests/ansible/Dockerfile
+++ b/config/Dockerfiles/package_tests/ansible/Dockerfile
@@ -1,0 +1,32 @@
+FROM fedora:25
+LABEL maintainer "https://github.com/CentOS-PaaS-SIG/ci-pipeline"
+LABEL description="This container is meant to \
+use upstreamfirst tests to test packages, \
+provided a package name and an ansible inventory."
+
+# Get repo for restraint-rhts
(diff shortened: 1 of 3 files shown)


config/Dockerfiles/package_tests/ansible/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+LABEL maintainer "https://github.com/CentOS-PaaS-SIG/ci-pipeline"
+LABEL description="This container is meant to \
+use upstreamfirst tests to test packages, \
+provided a package name and an ansible inventory."
+
+# Get repo for restraint-rhts
+ADD https://copr.fedorainfracloud.org/coprs/bpeck/restraint/repo/fedora-25/bpeck-restraint-fedora-25.repo /etc/yum.repos.d/
+
+# Install all package requirements
+RUN for i in {1..5} ; do dnf -y install ansible \
```
scoheb: what happens if this fails every time and the loop finishes?

PR #186, opened 2017-07-07
Add docker container to run package level testing against an ansible host

johnbieren:
This container takes in an inventory file and a private key as inputs and runs package level testing against the ansible host.  It is set up to require both the ssh private key and the inventory file to be in the same directory on the host it is called from.  It also writes the artifacts out to that same directory.

Also fixes a small check in the rpmbuild task for deciding whether or not to build the image that Bill pointed out.

### DIFF
--- a/config/Dockerfiles/package_tests/ansible/ansible_package_test.sh
+++ b/config/Dockerfiles/package_tests/ansible/ansible_package_test.sh
@@ -10,11 +10,9 @@ if [ $? -ne 0 ]; then
      echo "No upstream repo for this package! Exiting..."
      exit 1
 fi
-# Clone jbieren fork of beakerlib role
-# Note: Once PR merged upstream, can just use upstream master
-git clone https://pagure.io/forks/jbieren/standard-test-roles.git
+# Clone standard-test-roles repo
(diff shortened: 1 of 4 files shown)


config/Dockerfiles/package_tests/ansible/ansible_package_test.sh
```diff
@@ -0,0 +1,36 @@
...
+cat << EOF > test_atomic.yml
+---
+- hosts: all
+  roles:
+  - role: standard-test-beakerlib
+    tests:
+EOF
+# Find the tests
+git clone https://upstreamfirst.fedorainfracloud.org/${package}
+for test in $(find ${package} -name "runtest.sh"); do
```
arilivigni: Should there be an if here if there are no runtest.sh files?  Then we exit?

johnbieren:
@arilivigni Good idea.  I added a check
@scoheb I believe what would happen is the docker build would fail and it would exit.  The reason why I added that originally is because every now and then, even though the mirrors were up, I would see a dnf install miss a mirror and the whole process would exit when if it had simply retried it would have succeeded.  So, my goal here is to only break if the mirrors are actually broken.

PR #187, opened 2017-07-10
Allow rsync_opts to be passed to the rsync container, but default to --delete

johnbieren:

### DIFF
--- a/config/Dockerfiles/rsync/rsync.sh
+++ b/config/Dockerfiles/rsync/rsync.sh
@@ -6,6 +6,10 @@ set -x
 #rsync_from="fedora-atomic@artifacts.ci.centos.org::fedora-atomic/f26/"
 #rsync_to="/home/output/"
 
+if [[ ! -v rsync_opts ]]; then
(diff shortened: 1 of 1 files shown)


config/Dockerfiles/rsync/rsync.sh
```diff
@@ -6,6 +6,10 @@ set -x
 #rsync_from="fedora-atomic@artifacts.ci.centos.org::fedora-atomic/f26/"
 #rsync_to="/home/output/"
 
+if [[ ! -v rsync_opts ]]; then
```
arilivigni: Would this simplify your line
rsync_opts=${rsync_opts:-"--delete"}

johnbieren: Absolutely.  Didn't know about that. Thanks!

johnbieren: Actually, that doesn't work if rsync_opts="" which is defined but empty, which is how I plan to use the variable when I don't want --delete

PR #191, opened 2017-07-10
Make the rpmbuild task more containerized

johnbieren:
Granted it is a lot of docker runs, this is a more containerized way to run the rpmbuild task.  Mostly I just containerized the rsync steps.  The locking is still commented out.

### DIFF
--- a/tasks/rpmbuild-test
+++ b/tasks/rpmbuild-test
@@ -1,69 +1,86 @@
 #!/bin/bash
 set +e
 
-base_dir="$(dirname $0)/.."
+function acquire_lock {
+     # Write uuid to a lock file and store a backup
+     uuidgen > file.lock
+     cp file.lock uuid.saved
+     while true; do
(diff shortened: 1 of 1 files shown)


tasks/rpmbuild-test
```diff
@@ -1,69 +1,86 @@
...
+          # If uuid matches, we can proceed
+          if [[ $(diff file.lock uuid.saved) == "" ]]; then
+               break
+          fi
+          sleep 60
+     done
+}
+
+# Kill backgrounded jobs on exit
+function clean_up {
```
p3ck: nothing calls this

johnbieren: Yea that's the point for now.  Nothing calls acquire_lock either.  We are leaving it commented out until everything is all set with the 2.0 pipeline

p3ck: ok

PR #210, opened 2017-07-24
Fix log reporting in ostree image compose.

p3ck:


config/Dockerfiles/ostree_image_compose/ostree-image-compose.sh
```diff
@@ -23,12 +23,14 @@ virtlogd &
 chmod 666 /dev/kvm
 
 function clean_up {
+  # Add some debug here since I want to make sure I see what's in $basedir
+  ls $base_dir
```
scoheb: Are you not sure what base_dir is or defined to?

I have had success with this:

`base_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"`

PR #218, opened 2017-07-24
Fix issue with the logs directory not being present

arilivigni:
- This directory doesn't exist on the duffy hose
- Link to output directory

### DIFF
--- a/tasks/ostree-image-compose
+++ b/tasks/ostree-image-compose
@@ -10,6 +10,7 @@ base_dir="$(dirname $0)/.."
 HOMEDIR=$(pwd)
 rm -rf ${HOMEDIR}/output
 mkdir -p ${HOMEDIR}/output
+ln -fs ${HOMEDIR}/output ${HOMEDIR}/logs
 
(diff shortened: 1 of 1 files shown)


arilivigni:
I know I am supposed to be on PTO but it is raining up here and we are all inside the cabins ;)

tasks/ostree-image-compose
```diff
@@ -10,6 +10,7 @@ base_dir="$(dirname $0)/.."
 HOMEDIR=$(pwd)
 rm -rf ${HOMEDIR}/output
 mkdir -p ${HOMEDIR}/output
+ln -fs ${HOMEDIR}/output ${HOMEDIR}/logs
```
p3ck: line 32 should mean this isn't required.

arilivigni: weird I thought it ran with your change and failed for the same reason when I went to the node for that stage name logs was not there.

arilivigni: Actually probably should have made this ln -fs ${HOMEDIR}/output/logs ${HOMEDIR}/logs

PR #220, opened 2017-07-24
Skip creating images atm

arilivigni:
- Skip generating images since it causes issues
- Also add stage message to checkLatestImage

Once this is split into libs for each stage
this will make more sense

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -256,7 +256,7 @@ node('fedora-atomic') {
                             "DUFFY_HOST=${env.DUFFY_HOST}"
 
                     // Check if a new ostree image compose is needed
-                    checkLastImage()
+                    //checkLastImage()
                     if (fileExists("${env.WORKSPACE}/NeedNewImage.txt")) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -738,7 +738,9 @@ def rsyncResults(stage) {
     }
 }
 
-def checkLastImage() {
+def checkLastImage(stage) {
```
scoheb: Why not return true/false instead of touching a file?

arilivigni: I just took what was there before that touched a file.  In the shell step how do you return that to the method being called.

scoheb: Here is a snippet of how to do it.
https://paste.fedoraproject.org/paste/iAtLDDDTMg6POXx3vvvO1A
maybe we can tackle it as part of a refactoring PR if you prefer.

arilivigni: I like it.  Sounds like it can be done as the massive refactoring. 
Libraries , sanitizing messages, etc.

scoheb:
+1

PR #229, opened 2017-07-25
Don't escape the variables.  We need the current values.

p3ck:
Also fix rsyncResults to call the tasks consistently and without a
double ././


arilivigni:
Looks like this causes issue with Groovy which is why I had it escaped.

PR #244, opened 2017-08-02
Force generation of the atomic host image

arilivigni:
- Added boolean GENERATE_IMAGE
- Also archive the trigger.downstream file


Jenkinsfile
```diff
@@ -257,7 +258,7 @@ node('fedora-atomic') {
 
                     // Check if a new ostree image compose is needed
                     checkLastImage("${current_stage}")
-                    if (fileExists("${env.WORKSPACE}/NeedNewImage.txt")) {
+                    if (fileExists("${env.WORKSPACE}/NeedNewImage.txt") || ("${env.GENERATE_IMAGE}" == "true")) {
                         stage('ci-pipeline-ostree-image-compose') {
```
scoheb: maybe this if should be inside the stage? That way when looking at stage view we will also see the correct numbers of stages?

arilivigni: Good point I will make an update

PR #250, opened 2017-08-07
Add package level testing to ostree-boot task for now

johnbieren:
We will want this to be its own job.  However, I think we should turn this on for now until we get there, so we can say we are running these tests in the pipeline.  In order for it to be its own job, we will have to modify the ostree-boot task to not tear down the host and to pass its inventory file and key along to the package testing task.  This would require not tearing down the duffy host as well.  But, we are moving to containers, so not sure what it means in that case.  So, since adding it cleanly is a little complicated for now, I think we should temporarily add it like this.  Once we have ostree-boot working in a container, we can worry about separating it out and defining the workflow for how the two work together.

### DIFF
--- a/jobs/ostree-compose.yml
+++ b/jobs/ostree-compose.yml
@@ -243,6 +243,7 @@
             task: ostree-boot-image
             variables: |
                 export branch="${fed_branch}"
+                export fed_repo="${fed_repo}"
                 export image2boot="${image2boot:-}"
                 export commit=${commit:-}
                 export JENKINS_JOB_NAME="${JOB_NAME}"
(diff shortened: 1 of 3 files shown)


jobs/ostree-compose.yml
```diff
@@ -243,6 +243,7 @@
             task: ostree-boot-image
             variables: |
                 export branch="${fed_branch}"
+                export fed_repo="${fed_repo}"
```
arilivigni: Is this to make sure we are up to date since we don't run this any longer?

johnbieren: Oh yea... my bad. Let me add it to the Jenkinsfile too

PR #254, opened 2017-08-09
Move to using withEnv()

scoheb:


JenkinsfileTrigger
```diff
@@ -64,84 +67,87 @@ node('master') {
...
-                                        "status=${currentBuild.currentResult}"
-                    messageContent = ''
-
-                    // Send message org.centos.prod.ci.pipeline.package.queued or .ignore on fedmsg
-                    sendMessage(messageProperties, messageContent)
+                            exit
+                        '''
+                    }
+
+                    def jobproperties_map = readProperties file: "{env.WORKSPACE}/job.properties"
```
arilivigni: I don't think we need this for the message properties we send we need it for shell files only that is the time when we do the conversion

PR #256, opened 2017-08-09
withEnv does not work as expected

arilivigni:
- needs some testing reverting the change


arilivigni:
@scoheb this needs some more testing to validate I reverted the change since the pipeline is failing

PR #257, opened 2017-08-09
Improve rpmbuild container and have it do the rsync as well

johnbieren:
These changes to the rpmbuild container are meant to make the migration into OpenShift easier.  I put the rsync aspects of the rpmbuild flow into the container so it is all one container call.  It will still exit before going near any rsync stuff if make test fails or any of the rpmbuild stuff fails.  No changes should be needed in the Jenkinsfile, as no additional arguments are needed and it is still just a call to the duffy builder with the same task.  I made all the changes in the task file and container config so it should *just work.*  

Once approved, I will merge it when I am going to be online for a bit so I can monitor it to make sure I didn't break anything.

### DIFF
--- a/config/Dockerfiles/rpmbuild/Dockerfile
+++ b/config/Dockerfiles/rpmbuild/Dockerfile
@@ -1,7 +1,8 @@
 FROM fedora:24
 LABEL maintainer "https://github.com/CentOS-PaaS-SIG/ci-pipeline"
 LABEL description="This container is meant to \
-use fedpkg mock to create rpms."
+use fedpkg mock to create rpms.  It also rsyncs \
+them to artifacts.ci.centos.org."
 
 # Install all package requirements
 RUN for i in {1..5} ; do dnf -y update && dnf clean all && break || sleep 10 ; done
@@ -39,3 +40,5 @@ COPY rpmbuild-test.sh /home/rpmbuild-test.sh
 
 # Run the build script
 ENTRYPOINT ["bash", "/home/rpmbuild-test.sh"]
+# Run the container as follows
+# docker run --privileged -v /some/spot/for/logs:/home/logs -e fed_repo=${packagename} -e fed_branch=${fed_branch} -e fed_rev=${fed_rev} -e RSYNC_PASSWORD=${rsync_password} container_tag
--- a/config/Dockerfiles/rpmbuild/Dockerfile
+++ b/config/Dockerfiles/rpmbuild/Dockerfile
@@ -41,4 +41,4 @@ COPY rpmbuild-test.sh /home/rpmbuild-test.sh
 # Run the build script
 ENTRYPOINT ["bash", "/home/rpmbuild-test.sh"]
 # Run the container as follows
-# docker run --privileged -v /some/spot/for/logs:/home/logs -e fed_repo=${packagename} -e fed_branch=${fed_branch} -e fed_rev=${fed_rev} -e RSYNC_PASSWORD=${rsync_password} container_tag
+# docker run --privileged -v /some/spot/for/logs:/home/logs -e fed_repo=${packagename} -e fed_branch=${fed_branch} -e fed_rev=${fed_rev} -e enable_rsync=yes -e RSYNC_PASSWORD=${rsync_password} -e production=yes container_tag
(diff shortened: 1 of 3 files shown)


config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -1,74 +1,117 @@
 #!/bin/bash
 set +e
 
-OUTPUTDIR=/home/${fed_repo}/output
-which fedpkg
-if [ "$?" != 0 ]; then echo "ERROR: FEDPKG RPM NOT INSTALLED\nSTATUS: $?"; exit 1; fi
-# Put all output files into logs/ for rsync
-rm -rf ${OUTPUTDIR}/logs
-mkdir ${OUTPUTDIR}/logs
+RPMDIR=/home/${fed_repo}_repo
```
scoheb: Should we include some sort of validation that we have all the input env vars?

`if [ - z "${fed_repo}"] ; then
  echo "no fed_repo env var"
  exit 1
fi
`

config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -1,74 +1,117 @@
...
+    # Pull lock file back
+     rsync -avz fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}/repo/lockdir/file.lock file.lock
+    # If uuid matches, we can proceed
+     if [[ $(diff file.lock uuid.saved) == "" ]]; then
+          break
+     fi
+     sleep 60
+done
+# Rsync the empty directories over first, then the repo directory
+rsync -arv ${RSYNC_BRANCH}/ fedora-atomic@artifacts.ci.centos.org::fedora-atomic
```
scoheb: should we also add a way to disable rsync for future testing? Let's say we want to validate building a package without doing the rsync?

arilivigni: I think we need a stage pipeline to test changes and not do rsync there since that should be a known quantity IMHO

config/Dockerfiles/rpmbuild/rpmbuild-test.sh
johnbieren: I'll do that

config/Dockerfiles/rpmbuild/rpmbuild-test.sh
johnbieren: I can make it an argument to the container run. Good suggestion

johnbieren:
I have made modifications based on your suggestions, Ari and Scott, I will not be merging this until the staging pipeline is ready, though.

config/Dockerfiles/rpmbuild/Dockerfile
```diff
@@ -41,4 +41,5 @@ COPY rpmbuild-test.sh /home/rpmbuild-test.sh
 # Run the build script
 ENTRYPOINT ["bash", "/home/rpmbuild-test.sh"]
 # Run the container as follows
+# Note: For stage builds, don't do production=no, simply don't pass a production var in
```
scoheb: should this comment read "don't do production=yes" or "do production=no"?

config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -82,7 +82,10 @@ RSYNC_BRANCH=${fed_branch}
 if [ "${fed_branch}" = "master" ]; then
     RSYNC_BRANCH=rawhide
 fi
-mkdir ${RSYNC_BRANCH}
+if [ -z "${production}" ]; then
```
scoheb: if [ "${production}" != "yes" ] ?

config/Dockerfiles/rpmbuild/Dockerfile
johnbieren: My intent was to just have it set so that if you don't pass in a production environment variable, you are assumed to not be production.  This way, production=yes production=true production=YES production=please all work

scoheb: I guess the comment "production=no" threw me off...since it will be production even though you said no :)

But I think in practice, we are going to have a job parameter that will be production_mode. And in our Jenkinsfile we will startup the script with:

`production=${production_mode}`

The way the code is now...we will need an extra if statement to determine if we need to pass in the env var.

PR #264, opened 2017-08-10
Inline the definition of pod and container templates

scoheb:
- This lays the groundwork for adding containers for testing


scoheb:
The diff is ugly...but I just added a section around our `node()`

PR #266, opened 2017-08-10
Handle dashes in FedMsg fields

BillRainford:
Per my conversation with Ari, creating a PR so that Upstream CI doesn't hit this issue if some future field in a message they receive also has a dash the way the message downstream CI gets does. (this is to address how we get 'message-content' and have to change it to 'message_content' to make it serializable.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,8 +102,8 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud: 
                         if [ "${branch}" = "master" ]; then
                           branch="rawhide"
                         fi
-
-
+                        
+                        
                         # Save the bramch in job.properties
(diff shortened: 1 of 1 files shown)


arilivigni:
I think the only changes should be the lines for the replace please clean-up white space and extra lines

BillRainford:
Hi Guys, 

On downstream CI for all our repos we have an automated gate on checkins to make sure we adhere to a coding standard (PEP8 for Python etc) and with that we all setup Atom to do a certain amount of checking/enforcing while we are editing to keep the gatekeeper happy. The whitepace changes were automagically added as we don't allow empty lines with whitespace which is why I didn't see that until I got the diff from the PR.  I made a second checkin to put back the whitespace onto your otherwise empty lines and the diff shows the two lines I changed for this PR. 
https://github.com/CentOS-PaaS-SIG/ci-pipeline/pull/266/files

Take care,
-Bill

PR #267, opened 2017-08-11
removing the output dir was really just to clear the logs

p3ck:
from the previous run.  So just delete the logs and it must be
done via sudo because the containers run privileged.

### DIFF
--- a/tasks/ostree-compose
+++ b/tasks/ostree-compose
@@ -10,7 +10,8 @@ HOMEDIR=$(pwd)
 # Make sure logs are empty so we don't report previous build
 # if we fail
 if [ -e "${HOMEDIR}/output/logs" ]; then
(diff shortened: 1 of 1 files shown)


tasks/ostree-compose
```diff
@@ -6,7 +6,12 @@ fi
 
 base_dir="$(dirname $0)/.."
 HOMEDIR=$(pwd)
-rm -rf ${HOMEDIR}/output
+
+# Make sure logs are empty so we don't report previous build
+# if we fail
+if [ -e "${HOMEDIR}/output/logs" ]; then
+    sudo rm -rf ${HOMEDIR}/output/logs
```
scoheb: do we need to send this to /dev/null or a file in case?

p3ck: It's the previous builds logs so sending them to /dev/null is fine.

PR #270, opened 2017-08-14
Move latest symlink logic

p3ck:
Only update the symlink when the boot sanity job passes.
also make sure we don't garbage collect what the latest symlink is pointing to.

### DIFF
--- a/config/Dockerfiles/ostree_image_compose/ostree-image-compose.sh
+++ b/config/Dockerfiles/ostree_image_compose/ostree-image-compose.sh
@@ -106,7 +106,7 @@ if [ -e "latest-atomic.qcow2" ]; then
 fi
 
 # delete images over 3 days old but don't delete what our latest link points to
-find . -type f -mtime +3 ! -name "$latest" -exec rm {} \;
+find . -type f -mtime +3 ! -name "$latest" -exec rm -v {} \;
 
 # Also delete if we have more than 3 images (from manual runs)
 # but don't delete what our latest link points to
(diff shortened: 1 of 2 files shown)


config/Dockerfiles/ostree_image_compose/ostree-image-compose.sh
```diff
@@ -99,13 +99,22 @@ imagefactory --debug --imgdir $imgdir --timeout 3000 base_image /home/output/log
...
-# delete images over 3 days old
-find /home/output/images/ -type f -mtime +3 -exec rm {} \;
+pushd /home/output/images || exit 1
+latest=""
+if [ -e "latest-atomic.qcow2" ]; then
+    latest=$(readlink latest-atomic.qcow2)
+fi
+
+# delete images over 3 days old but don't delete what our latest link points to
+find . -type f -mtime +3 ! -name "$latest" -exec rm {} \;
```
scoheb: if these items are just files, should we add '-v' to rm to see what is it doing?

config/Dockerfiles/ostree_image_compose/ostree-image-compose.sh
```diff
@@ -99,13 +99,22 @@ imagefactory --debug --imgdir $imgdir --timeout 3000 base_image /home/output/log
...
 # Also delete if we have more than 3 images (from manual runs)
-pushd /home/output/images
-ls -t | sed -e '1,3d' | xargs --no-run-if-empty -d '\n' rm
-ln -sf $imgname.qcow2 latest-atomic.qcow2
+# but don't delete what our latest link points to
+images=$(ls -t)
+if [ -n "$latest" ]; then
+    images=$(echo $images| grep -v "$latest")
+fi
+echo $images| sed -e '1,3d' | xargs --no-run-if-empty -d '\n' rm
```
scoheb: if these items are just files, should we add '-v' to rm to see what is it doing?

p3ck: good suggestion.

PR #277, opened 2017-08-15
package_list: add rpm-ostree

jlebon:
rpm-ostree has a nice integration testsuite which would be great to run
in the CI pipeline. Add it to the package list.


arilivigni:
I assume this is a core of Atomic Host?

PR #280, opened 2017-08-16
trigger job: add f27

jlebon:
We were looking for `master` but comparing it to `rawhide`. Also add f27
now that dist-git branched.


jlebon:
Nevermind the original description, I misread the output. This PR is now just about adding f27.

arilivigni:
This is our old pipeline not what we run today and we are also limiting this to f26 for now

PR #281, opened 2017-08-16
Add openshiftVerifyBuild for rpmbuild

robnester-rh:
Added the openshiftVerifyBuild for a minimal build validation.

From my discussion with @arilivigni I believe this is a minimum validation requirement for validating our build.


scoheb:
not sure we need this anymore?

arilivigni:
Not needed anymore

PR #283, opened 2017-08-17
Add rpmbuild container call

scoheb:
- replace duffy calls with container step for rpmbuild stage
- refactor to allow pipeline to be parameterized for provisioning

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,8 @@ properties(
                                 string(defaultValue: 'f26', description: 'Fedora target branch', name: 'TARGET_BRANCH'),
                                 string(defaultValue: 'ci-pipeline', description: 'Main project repo', name: 'PROJECT_REPO'),
                                 string(defaultValue: 'org.centos.stage', description: 'Main topic to publish on', name: 'MAIN_TOPIC'),
+                                string(defaultValue: 'production=yes', description: 'Production Mode. Leave empty to disable.', name: 'PRODUCTION_MODE'),
+                                string(defaultValue: 'enable_rsync=yes', description: 'Rsync Mode. Leave empty to disable rsync.', name: 'RSYNC_MODE'),
                                 string(defaultValue: 'fedora-fedmsg', description: 'Main provider to send messages on', name: 'MSG_PROVIDER'),
                                 string(defaultValue: 'stable', description: 'Tag for rpmbuild image', name: 'RPMBUILD_TAG'),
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -139,48 +151,20 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud:
...
-                        writeFile file: "${env.ORIGIN_WORKSPACE}/task.env",
-                                text: "export JENKINS_JOB_NAME=\"${JOB_NAME}-${current_stage}\"\n" +
-                                        "export JENKINS_BUILD_TAG=\"${BUILD_TAG}-${current_stage}\"\n" +
-                                        "export OSTREE_BRANCH=\"${OSTREE_BRANCH}\"\n" +
-                                        "export fed_repo=\"${fed_repo}\"\n" +
-                                        "export fed_branch=\"${fed_branch}\"\n" +
-                                        "export fed_rev=\"${fed_rev}\"\n"
-                        rsyncResults("${current_stage}")
-
-                        def package_props = "${env.ORIGIN_WORKSPACE}/logs/package_props.txt"
```
p3ck: I assume there are some properties from the rpm build container that need to be exposed to the next stage?  How does that work in this model?

scoheb: nice catch. Yes, I presume these props need to be passed...I will add them back

Jenkinsfile
```diff
@@ -139,48 +151,25 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud:
...
-
-                        if (env.OSTREE_BRANCH == null) {
-                            env.OSTREE_BRANCH = ""
+                        // Call rpmbuild-test.sh
+                        // in rpmbuild container
+                        container('rpmbuild') {
+                            sh "fed_repo=${env.fed_repo} " +
+                               "fed_branch=${env.fed_branch} " +
+                               "fed_rev=${env.fed_rev} " +
+                               "/home/rpmbuild-test.sh"
```
johnbieren: We will also need RSYNC_PASSWORD, enable_rsync=yes and production=yes (my apologies if they are in here somewhere else and I missed them)

arilivigni:
Can we not make this change until I have the libs in place otherwise I have to keep playing catchup.  I will have an initial review shortly and we can make changes but I don't think this is a blocker is it?

scoheb:
For sure...we can wait...we actually need the staging env first before we merge this I would think too...

arilivigni:
Sounds good thanks!

scoheb:
Closed in favor of https://github.com/CentOS-PaaS-SIG/ci-pipeline/pull/348

PR #288, opened 2017-08-18
More substitution needed in fedmsg_fields.txt

arilivigni:
- This caused a recent failure
  https://jenkins-continuous-infra.apps.ci.centos.org/job/continuous-infra-ci-pipeline-f26/36/console
- We should just throw out anything that doesn't start with
  ^fed since we don't need this issue

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -84,6 +84,7 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud: 
                         if [ -n "${CI_MESSAGE}" ]; then
                             echo ${CI_MESSAGE} | ${WORKSPACE}/parse_fedmsg.py > fedmsg_fields.txt
                             sed -i '/^\\\\s*$/d' ${WORKSPACE}/fedmsg_fields.txt
+                            sed -i '/`/g' ${WORKSPACE}/fedmsg_fields.txt
                             sed -i '/^fed/!d' ${WORKSPACE}/fedmsg_fields.txt
                             grep fed ${WORKSPACE}/fedmsg_fields.txt > ${WORKSPACE}/fedmsg_fields.txt.tmp
                             mv ${WORKSPACE}/fedmsg_fields.txt.tmp ${WORKSPACE}/fedmsg_fields.txt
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -85,6 +85,7 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud:
                             echo ${CI_MESSAGE} | ${WORKSPACE}/parse_fedmsg.py > fedmsg_fields.txt
                             sed -i '/^\\\\s*$/d' ${WORKSPACE}/fedmsg_fields.txt
                             sed -i '/`/g' ${WORKSPACE}/fedmsg_fields.txt
+                            sed -i '/^fed/!d\' ${WORKSPACE}/fedmsg_fields.txt
```
p3ck: Is the intent to delete all fields which don't start with fed?  Also - why is the closing quote escaped?  I don't think that will work,

scoheb: @p3ck  is right:

> shebert@localhost/tmp:$ cat fed.txt
> fed_scott=tom
> redhat_go=123
> fed_bef=abc
> shebert@localhost/tmp:$ 
> shebert@localhost/tmp:$ sed -i '/^fed/!d\' fed.txt 
> sed: -e expression #1, char 9: extra characters after command
> shebert@localhost/tmp:$ sed -i '/^fed/!d' fed.txt 
> shebert@localhost/tmp:$ cat fed.txt
> fed_scott=tom
> fed_bef=abc
> shebert@localhost/tmp:$ 
>

arilivigni: new update

Jenkinsfile
```diff
@@ -84,7 +84,7 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud:
                         if [ -n "${CI_MESSAGE}" ]; then
                             echo ${CI_MESSAGE} | ${WORKSPACE}/parse_fedmsg.py > fedmsg_fields.txt
                             sed -i '/^\\\\s*$/d' ${WORKSPACE}/fedmsg_fields.txt
```
scoheb: Is this supposed to get rid of empty lines?

arilivigni: Probably that was there already but that isn't the issue that I fixed

PR #299, opened 2017-08-22
Drop arch.rpm from the original_spec_nvr and the .rpm from the nvr field (in the doc)

pypingou:


pypingou:
I don't think the ``.rpm`` ever was included:
````
$ rpm -q --define "dist .rawhide" --queryformat '%{name}-%{version}-%{release}.%{arch}\n' --specfile guake.spec
````

Regarding the nvr field I wonder if the current approach works as expected, see for example: https://apps.stg.fedoraproject.org/datagrepper/id?id=2017-56517bde-f80e-411a-baa9-c7b836623dee&is_raw=true&size=extra-large where the field is empty

pypingou:
Oh I see the NVR issue is fixed in #300 cool :)

pypingou:
I adjust the nvr field in the doc assuming it was the same as original_spec_nvr but if that's not the case I can just drop the second commit.

> Not 100% sure why we want arch on one of the fields and not the other.

I'm currently only relying on original_spec_nvr but I'm fine with adjusting the nvr field in the same way (or undo my change to the README).

arilivigni:
@pypingou Thanks for the PR I think we should drop arch for both and documentation should reflect that as well.  Can you make that update please?

pypingou:
Will do :)

pypingou:
Updated and rebased :)

PR #303, opened 2017-08-23
Revamp rpmbuild container

johnbieren:
This is again the PR to put everything including all the rsync stuff inside the container run call.  This has already been reviewed and approved by multiple people.  The only new thing here is line 14 of the task file. I was previously mounting ${HOMEDIR}:/home/logs which was putting all the log files in ${HOMEDIR}, but not creating a log dir in ${HOMEDIR}, which is what the duffy publisher looks for, thus it was failing on just that.  So, as long as my new mount thing makes sense, you shouldn't have to review the rest as it has already been reviewed and we should be good to go.


johnbieren:
Now that the koji credentials are live I will uncomment out that stuff and turn on concurrent koji building, but I want to do that as a separate PR. So, hoping to get this merged and see it working first.

config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -1,79 +1,146 @@
...
+    # If uuid matches, we can proceed
+     if [[ $(diff file.lock uuid.saved) == "" ]]; then
+          break
+     fi
+     sleep 60
+done
+# Rsync the empty directories over first, then the repo directory
+rsync -arv ${RSYNC_BRANCH}/ fedora-atomic@artifacts.ci.centos.org::fedora-atomic
+if [ -z "${production}" ]; then
+     rsync --delete --stats -a ${RPMDIR} fedora-atomic@artifacts.ci.centos.org::fedora-atomic/${RSYNC_BRANCH}
```
p3ck: I don't get this.  
if it's not production it's going to delete everything under $RSYNC_BRANCH and replace it with $RPMDIR?

this should be using the variables I defined at the global level anyway.  Then this $production variable can go away.
$RSYNC_USER, $RSYNC_SERVER and $RSYNC_DIR

johnbieren: The intent was that yes, because there was discussion of for the staging branch, to just keep one directory of rpms in the same place for just one package at a time, so it would clear the old rpms first.  However, I have updated the container now to use RSYNC_USER, RSYNC_SERVER and RSYNC_DIR, so everything should be much cleaner now.

config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -1,79 +1,138 @@
...
 RPM_TO_CHECK=$(find /${fed_repo}/results_${fed_repo}/${VERSION}/*/ -name "${fed_repo}-${VERSION}*" | grep -v src)
-libabigail/tools/fedabipkgdiff --from ${ABIGAIL_BRANCH} ${RPM_TO_CHECK} &> ${OUTPUTDIR}/logs/fedabipkgdiff_out.txt
+libabigail/tools/fedabipkgdiff --from ${ABIGAIL_BRANCH} ${RPM_TO_CHECK} &> ${LOGDIR}/fedabipkgdiff_out.txt
 RPM_NAME=$(basename $RPM_TO_CHECK)
-echo "package_url=${HTTP_BASE}/${fed_branch}/repo/${fed_repo}_repo/$RPM_NAME" >> ${OUTPUTDIR}/logs/package_props.txt
+echo "package_url=${HTTP_BASE}/${fed_branch}/repo/${fed_repo}_repo/$RPM_NAME" >> ${LOGDIR}/package_props.txt
 echo "original_spec_nvr=${truenvr}" >> ${OUTPUTDIR}/logs/package_props.txt
 RPM_NAME=$(echo $RPM_NAME | rev | cut -d '.' -f 2- | rev)
 echo "nvr=${RPM_NAME}" >> ${OUTPUTDIR}/logs/package_props.txt
+RSYNC_LOCATION="${RSYNC_HOST}::${RSYNC_DIR}/${RSYNC_BRANCH}"
```
p3ck: I think it's RSYNC_SERVER isn't it?

johnbieren: It is, but I did -e RSYNC_HOST="${RSYNC_USER}@${RSYNC_SERVER}"  I can switch it to pass each of those to my container independently for consistency's sake if you think that is wiser.

p3ck: It just confused me. Which doesn't take much these days. I'll leave it up to you.

arilivigni: I defer to you if you want to change it.  I assume all the change have the other merged code with original_spec_nvr and nvr

johnbieren: Yea, I redid it with all the original_spec_nvr and nvr stuff.  I will merge it in the morning when I will be on for a while so I can monitor it again to double check. Will also be working on getting the concurrent koji build in as well.

PR #308, opened 2017-08-24
UPDATE python file to parse and write groovy props

robnester-rh:
Updated the writeFile call with a python script that does the following:

* Parses the CI_MESSAGE env variable into JSON
* If the CI_MESSAGE contains a 'commit' key, it performs the following
  * For any key, if the key value is a string, replaces any double quote
    character within the value with a single quote.
  * Truncates the value of the message key at the first newline character
  * Writes the key/values in the following format to a file named
    fedmsg_fields.groovy in the workspace.
    * env.fed_{key}="{value}"

This removes the need to convert any files and this fedmsg_fields.groovy
file is injected with a call to load the file object.


scoheb:
not a python expert at all...but lgtm

PR #311, opened 2017-08-24
UPDATE python script in Jenkinsfile

robnester-rh:
Updated python script in Jenkinsfile to reflect changes made in
rpmBuild.groovy

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -96,9 +96,7 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud: 
                         sh '${WORKSPACE}/parse_fedmsg.py'
 
                         // Load fedmsg fields as environment variables
-                        def fedmsg_fields = "${env.WORKSPACE}/fedmsg_fields.txt"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -71,36 +71,29 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud:
...
+                                        "            f.write('env.fed_{0}=\"{1}\"\\n'.format(k.replace('-', '_', ci_message[k]))"
+
+                        // Chmod the python script to make it executable
+                        sh 'chmod +x ${WORKSPACE}/parse_fedmsg.py'
+
+                        // Execute the python script
+                        sh '${WORKSPACE}/parse_fedmsg.py'
 
                         // Load fedmsg fields as environment variables
                         def fedmsg_fields = "${env.WORKSPACE}/fedmsg_fields.txt"
```
arilivigni: How does this get generated?

robnester-rh: The .txt file reference has now been removed. Instead we def fedmsg_fields_groovy pointing to the groovy file the python script is writing out, and then load that.

PR #321, opened 2017-08-28
Check if there are atomic tagged tests first

johnbieren:
Currently the package level test containers just call the playbook if there are tests there that are beakerlib.  If there are beakerlib tests there, but they are not tagged for atomic, it will run and pass but have no results because nothing run, which is kind of confusing.  With this, it will echo out that there are no atomic tagged tests and do nothing.

### DIFF
--- a/config/Dockerfiles/package_tests/ansible/ansible_package_test.sh
+++ b/config/Dockerfiles/package_tests/ansible/ansible_package_test.sh
@@ -9,7 +9,7 @@
 curl -s --head https://upstreamfirst.fedorainfracloud.org/${package} | head -n 1 | grep "HTTP/1.[01] [23].." > /dev/null
 if [ $? -ne 0 ]; then
      echo "No upstream repo for this package! Exiting..."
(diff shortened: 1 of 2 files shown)


config/Dockerfiles/package_tests/ansible/ansible_package_test.sh
```diff
@@ -17,6 +17,10 @@ if [[ $(grep "standard-test-beakerlib" ${package}/*.yml) == "" ]]; then
         exit 0
 fi
 if [ -f ${package}/tests.yml ]; then
+     if [[ $(ansible-playbook --list-tags ${package}/tests.yml) != *"atomic"* ]]; then
```
scoheb: Have you tested if left-hand side returns a string with spaces?

johnbieren: I tried this with a real playbook both with and without the atomic tag and it worked for both so should be good

PR #323, opened 2017-08-29
Fix up rpmbuild Makefile find

johnbieren:
The rpmbuild container searches for the Makefile's directory.  This was dependent on said directory containing the package name in its name.  That is not always the case.  This improves on that.  Also, some fedpkg preps (vim in particular) require perl, so install that.

Manually calling make test and ./configure may not be the best idea, so will have to be revisited.

### DIFF
--- a/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
+++ b/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
@@ -38,20 +38,22 @@ fedpkg --release ${fed_branch} prep
 VERSION=$(rpmspec --queryformat "%{VERSION}\n" -q ${fed_repo}.spec | head -n 1)
 # Some packages are packagename-version-release, some packagename-sha, some packagename[0-9]
 DIR_TO_GO=$(dirname $(find . -name Makefile | tail -n 1))
-pushd $DIR_TO_GO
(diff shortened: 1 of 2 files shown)


config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -37,7 +37,7 @@ sed -i "/^Release:/s/%{?dist}/.${commits}.${fed_rev:0:7}%{?dist}/" ${fed_repo}.s
 fedpkg --release ${fed_branch} prep
 VERSION=$(rpmspec --queryformat "%{VERSION}\n" -q ${fed_repo}.spec | head -n 1)
 # Some packages are packagename-version-release, some packagename-sha, some packagename[0-9]
-DIR_TO_GO=$(find . -maxdepth 1 -type d | cut -c 3- | grep ${fed_repo})
+DIR_TO_GO=$(dirname $(find . -name Makefile | tail -n 1))
```
p3ck: I think I would still wrap this in a if [ -n "$DIR_TO_GO" ]; then....

johnbieren: I was considering that, I'll push another commit with that change

PR #325, opened 2017-08-29
Debugging atomic-host-tests stage

johnbieren:
The atomic-host-tests stage is provisioning duffy, exiting, then tearing down the duffy instance.  Not sure if this is why, but it is my first guess.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -602,6 +602,7 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud: 
                         // Set groovy and env vars
                         env.task = "./ci-pipeline/tasks/atomic-host-tests"
                         env.playbook = "ci-pipeline/playbooks/system-setup.yml"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -602,6 +602,7 @@ podTemplate(name: 'fedora-atomic-inline', label: 'fedora-atomic-inline', cloud:
                         // Set groovy and env vars
                         env.task = "./ci-pipeline/tasks/atomic-host-tests"
                         env.playbook = "ci-pipeline/playbooks/system-setup.yml"
+                        env.ANSIBLE_HOST_KEY_CHECKING = "False"
```
arilivigni: I believe this is set for the overall pipeline so it should not to be set again

johnbieren: It is set in rpmbuild, ostree-compose, ostree-image-compose, ostree-image-boot-sanity, and ostree-boot-sanity currently. Figured it was worth a try

PR #335, opened 2017-09-07
Change the directories and file we check for xUnit

arilivigni:
- **/logs/ansible_xunit.xml

This worked:
https://jenkins-continuous-infra.apps.ci.centos.org/job/continuous-infra-ci-pipeline-f26/190/


johnbieren:
Thanks Ari

PR #337, opened 2017-09-07
Better modification of .spec file's release

johnbieren:
[root@hp-z420-01 container-selinux]# cat container-selinux.spec | grep '%{?dist}'
Release: 1%{?dist}
[root@hp-z420-01 container-selinux]# cat ../glibc/glibc.spec | grep '%{?dist}'
%define glibcrelease 15%{?dist}
[root@hp-z420-01 container-selinux]#

Not every .spec file just says Release: foo%{?dist}. Some, like glibc shown above, use other variables, but they will use %{?dist} in their definition regardless, so just replace that.


scoheb:
not too familiar with this area...but LGTM

PR #340, opened 2017-09-07
Clear ~/.ssh/known_hosts every time

johnbieren:
16:33:04 Add correct host key in /home/jenkins/.ssh/known_hosts to get rid of this message.
16:33:04 Offending ECDSA key in /home/jenkins/.ssh/known_hosts:4

This should fix that error as far as I know.  Hopefully I put it in the right spot.


johnbieren:
Above check failure is not an actual problem with the PR, but due to the check job pointing to a separate Jenkinsfile at the moment

scoheb:
lgtm

PR #342, opened 2017-09-08
Revert removal of known_hosts

johnbieren:
The pipeline doesn't like connecting to duffy without a known_hosts file either.  We will need to do something with the UsersKnownHostsFile or something. Not sure yet.


johnbieren:
So we will still have problems with the pipeline trying to use a duffy hostname it used in the past, but removing known_hosts each time is somehow breaking it from connection to duffy at all

PR #347, opened 2017-09-08
branch and repo need to exist first for locking functions to work.

p3ck:


config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -110,6 +110,10 @@ if [ -z "${RSYNC_PASSWORD}" ]; then echo "Told to rsync but no RSYNC_PASSWORD en
 # Perform rsync to artifacts.ci.centos.org
 mkdir -p ${RSYNC_BRANCH}
 mkdir repo
+# Rsync the empty directories over first, then the repo directory
+rsync -arv ${RSYNC_BRANCH}/ ${RSYNC_HOST}::${RSYNC_DIR}
+rsync -arv repo/ ${RSYNC_LOCATION}
```
robnester-rh: This won't work as written. You need to add the -R flag, otherwise the files from repo are copied (if there are any) but not the repo dir itself.

robnester-rh: Additionally, if there are other scripts that create dirs under $RSYNC_DIR/$RSYNC_BRANCH you might want to ensure that they use the -R flag so that they create any given local dirs instead of just rsyncing the contained files.

johnbieren: Are you sure about that? Because every document I see that says how to copy over a remote dir does not have -R.  In fact, a document telling you how to clear out a remote dir but rsync'ing over $(mktemp -d) does not even have -R, which is exactly rsync'ing over solely a dir, as we are here.

robnester-rh: Further exploration of this as determined that the reason -R seemed to be necessary was due to the trailing '/' on repo. If we remove the trailing '/' on repo, then -arv does create the repo dir at the remote location as expected.

PR #352, opened 2017-09-12
whitelist check for PR author or keyword commenter

cevich:
Add two new options: --whitelist and --limit-whitelist

The first takes a file containing line-delimited github PR author names.
When specified, a PR will only be considered for triggering if the
author is in the referenced whitelist file.

The second option is related to --limit-keyword and requires --keyword.
When specified, it requires the keyword-commenter must be in the
whitelist file (the PR author is ignored).

Some other small changes were made to function-parameters to permit
simpler passing of options.  Also simplification of the triggering
conditional was made which will better support future addition of
conditions.

Signed-off-by: Chris Evich <cevich@redhat.com>

### DIFF
--- a/utils/pull_requests.py
+++ b/utils/pull_requests.py
@@ -2,20 +2,31 @@
 from __future__ import print_function
 import sys
 import time
-import requests
 import json
 from optparse import OptionParser
 import pprint
+import requests
 pp = pprint.PrettyPrinter(indent=4)
 token_header = dict()
 keyword = None
 
-
 def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
 
+_whitelist_cache = []
+def author_in_whitelist(author, whitelist_filename, pr_no):
+    if not _whitelist_cache:
+        with open(whitelist_filename, 'r') as whitelist:
+            for line in whitelist:
+                line = line.strip()
+                if line.startswith('#') or not line:
+                    continue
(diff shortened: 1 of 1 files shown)


cevich:
@johnbieren oh, thanks man 😀

utils/pull_requests.py
```diff
@@ -10,12 +10,18 @@
 token_header = dict()
 keyword = None
 
-
 def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
 
+def author_in_whitelist(author, options, pr_no):
```
p3ck: I would pass options.whitelist to this method instead of options since no other arg is used from options and it makes it clearer what the method is acting on.

cevich: TBH, that's how I had it before.  Then I started just passing ``options`` through to other functions, so I did it here for consistency-sake.  Though you're right, it's not strictly needed.  I'll remove it if that's a hard-requirement, otherwise I'm more comfortable with being lazy 😀

p3ck: it makes sense in the other areas where you are using multiple options.  Put it this way, if you were to put this in a library would you expect your callers to wrap that argument in an options object just so they can call your method?

cevich:
Fixed as requested, rebased against master and ``--force`` pushed.  Could use some help testing it though.

PR #355, opened 2017-09-18
Arch override koji builds

johnbieren:
Just do x86_64 koji builds

### DIFF
--- a/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
+++ b/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
@@ -64,7 +64,7 @@ rpmbuild -bs /root/rpmbuild/SOURCES/${fed_repo}.spec
 kinit -k -t /home/fedora.keytab $FEDORA_PRINCIPAL
 # Build the package into ./results_${fed_repo}/$VERSION/$RELEASE/ and concurrently do a koji build
 { time fedpkg --release ${fed_branch} mockbuild ; } 2> ${LOGDIR}/mockbuild.txt &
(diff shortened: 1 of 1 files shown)


config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -64,7 +64,7 @@ rpmbuild -bs /root/rpmbuild/SOURCES/${fed_repo}.spec
 kinit -k -t /home/fedora.keytab $FEDORA_PRINCIPAL
 # Build the package into ./results_${fed_repo}/$VERSION/$RELEASE/ and concurrently do a koji build
 { time fedpkg --release ${fed_branch} mockbuild ; } 2> ${LOGDIR}/mockbuild.txt &
-{ time koji build --wait --scratch $RSYNC_BRANCH /root/rpmbuild/SRPMS/${fed_repo}*.src.rpm ; } 2> ${LOGDIR}/kojibuildtime.txt &
+{ time koji build --wait --arch-override=x86_64 --scratch $RSYNC_BRANCH /root/rpmbuild/SRPMS/${fed_repo}*.src.rpm ; } 2> ${LOGDIR}/kojibuildtime.txt &
```
scoheb: is this something we can maybe refactor to be an env var or arg we can control for later?

p3ck: I think that is a good idea but I wouldn't hold up this review on it.

johnbieren: If the arch override fixes the timing thing and we decide to keep this here long term I can modify it to be cleaner as suggested

PR #359, opened 2017-09-19
Add groovy docstrings and documentation to methods.

robnester-rh:
Added groovy docstrings to methods to allow for automated doc
generation.

Added docs directory of automatically generated documentation
for the pipelineUtils class in vars/


scoheb:
how are the HTML files going to be generated? Will they remain static in our repo until someone updates them?

our stage job should have a step of validating the docs...

PR #366, opened 2017-09-21
Convert ostree_boot_image task to a container

p3ck:
The artifact.ci.centos.org forwarding should be done with a
proper http proxy instead.  But this works now.


tasks/ostree-boot-image
```diff
@@ -1,10 +1,23 @@
...
 env
 
 # Find the base dir
 base_dir="$(dirname $0)/.."
+HOMEDIR=$(pwd)
+
+function clean_up {
+    set +e
+    # Tear down the atomic host and image
+    sudo docker run --privileged -it --link libvirtd:libvirtd -e commit=${commit} -e image2boot=${image2boot} -e state=absent -v /var/lib/libvirt/:/var/lib/libvirt/ -v ${HOMEDIR}/output/:/root/output/ -v ${basedir}/:/root/ci-pipeline ostree_boot_image
```
johnbieren: So this is the part that I think will not be part of this stage and will instead come after the integration and package level tests once this is put into container() calls in the Jenkinsfile

config/Dockerfiles/ostree_boot_image/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY atomic7-testing.repo /etc/yum.repos.d
+COPY walters-buildtools.repo /etc/yum.repos.d
+
+RUN yum -y install libguestfs libguestfs-tools-c \
+                   git ostree rpm-ostree libvirt-client \
+                   PyYAML python2-setuptools virt-install \
+                   python-pip python-devel gcc net-tools \
+                   openssh-clients sshpass && yum clean all
+
+RUN pip install ara
```
johnbieren: Do you want to version lock this like it is here? https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/playbooks/roles/ara/tasks/main.yml

tasks/ostree-boot-image
p3ck: agreed.  But for now we can do it here.

config/Dockerfiles/ostree_boot_image/Dockerfile
p3ck: Running on the latest centos7.4 it won't work with the version locked at 13.

arilivigni: We discussed this and in the future the staging pipeline would catch this so it shouldn't be a concern not to pin it

PR #370, opened 2017-09-22
Add package level tests as a pipeline stage

johnbieren:
This PR Adds package tests from upstream first as a pipeline stage. It uses a duffy host to essentially just call containers, as ostree_boot_sanity does at time of writing.  It will be changed to container() calls after ostree_boot_sanity is, as it has a lot of the same logic.  The task file is essentially a copy of ostree_boot_sanity with a different runtest at the end.  The Dockerfile has a couple scripts being added with it right now... some of them not currently being used.  One is atomic-host-tests.sh, as that stage isn't set to use a container yet (to come), and the other is package-tests.sh (not tmp-).  I believe package-tests.sh will be close to what we end up calling when we move to container() calls so I did not remove it. Also modified some of the message topics to fit the specs in our README.  

### DIFF
--- a/config/Dockerfiles/rpmbuild/Dockerfile
+++ b/config/Dockerfiles/rpmbuild/Dockerfile
@@ -40,6 +40,12 @@ RUN echo "config_opts['plugin_conf']['lvm_root_opts']['size'] = '16G'" >> /etc/m
 RUN echo "config_opts['plugin_conf']['lvm_root_opts']['poolmetadatasize'] = '30G'" >> /etc/mock/site-defaults.cfg
 RUN echo "config_opts['basedir'] = '/home/rpmbuild/'" >> /etc/mock/site-defaults.cfg
 
+# Change the anongiturl for fedpkg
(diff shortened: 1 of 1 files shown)


config/Dockerfiles/runtest/Dockerfile_disabled
```diff
@@ -0,0 +1,46 @@
+FROM fedora:25
```
scoheb: what is this file for? it is named disabled...

johnbieren: It is the Dockerfile that I believe we will use once we switch to container() calls. I had to make a second Dockerfile that is live and contains all the libvirt provisioning stuff, but this container shouldn't need that functionality so I didn't want to just trash my previous work

scoheb: smart chap

johnbieren:
[retest]

johnbieren:
[test]

PR #377, opened 2017-09-26
Cleanup README and fix jenkinsfile issue with atomic-host-tests

johnbieren:
I believe the syntax error in the Jenkinsfile is messing with the stage name and causing it to not run a task file on the duffy host again.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -292,8 +292,7 @@ podTemplate(name: 'fedora-atomic-' + env.ghprbActualCommit,
                         pipelineUtils.sendMessage(messageFields['properties'], messageFields['content'])
                     }
                     currentStage = "ci-pipeline-atomic-host-tests"
-                    stage(
-                            currentStage) {
+                    stage(currentStage) {
                         pipelineUtils.setStageEnvVars(currentStage)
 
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -292,8 +292,7 @@ podTemplate(name: 'fedora-atomic-' + env.ghprbActualCommit,
                         pipelineUtils.sendMessage(messageFields['properties'], messageFields['content'])
                     }
                     currentStage = "ci-pipeline-atomic-host-tests"
-                    stage(
-                            currentStage) {
+                    stage(currentStage) {
```
scoheb: I fail to see how this was a syntax error?

johnbieren: This is the only one I see with a new line like this, and it is the only one not executing its duffy task.  It is pretty much a shot in the dark.  We had this problem before, and it was because there was no rsyncResults call in the atomic-host-tests stage.  The task isn't being run again, and I am not positive as to why.

johnbieren: Right, added commit to really fix it

scoheb:
the stage build was hung...running for 8 hrs...I killed it.

PR #378, opened 2017-09-26
Update README.md about queue messages

arilivigni:
- We don't have queue messages for ostree amd smoke test
- We do for package, functional tests, and integration tests


johnbieren:
I don't think we do for image, either.
https://github.com/CentOS-PaaS-SIG/ci-pipeline/pull/377

arilivigni:
Already in @johnbieren PR

PR #384, opened 2017-09-27
Add rpm build logs

johnbieren:
I believe adding build logs for the rpm stage should be this easy. I will be curious of the stage pipeline results on this though, as I am not 100% sure if the ostree stages will complain about something besides .rpm files being in the artifact directories.

### DIFF
--- a/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
+++ b/config/Dockerfiles/rpmbuild/rpmbuild-test.sh
@@ -101,6 +101,8 @@ cp ${CURRENTDIR}/${fed_repo}/results_${fed_repo}/${VERSION}/*/*.rpm ${RPMDIR}/
 # Run rpmlint
 rpmlint ${RPMDIR}/ > ${LOGDIR}/rpmlint_out.txt
 pushd ${RPMDIR} && createrepo .
+mkdir logs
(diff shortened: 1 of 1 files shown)


config/Dockerfiles/rpmbuild/rpmbuild-test.sh
```diff
@@ -101,6 +101,8 @@ cp ${CURRENTDIR}/${fed_repo}/results_${fed_repo}/${VERSION}/*/*.rpm ${RPMDIR}/
 # Run rpmlint
 rpmlint ${RPMDIR}/ > ${LOGDIR}/rpmlint_out.txt
 pushd ${RPMDIR} && createrepo .
+mkdir logs
```
scoheb: do we not already have a logs directory?

johnbieren: We do, but not in ${RPMDIR}. RPMDIR is the dir that gets rsynced to artifacts.ci, so right now it is just .rpms. This logs dir is not for archiving with the job run, but to get the logs up to the artifacts server so they are right next to the rpms. As far as I could tell, that is what we want. @arilivigni is that correct? Or was the goal just to have them in the jenkins build artifacts. I suppose I could put them there too...

arilivigni: Yes that is exactly right so when we rsync for an rpm or also a tree the build logs of what occurred is already there.

PR #386, opened 2017-09-28
tasks: Make ostree-boot-image runnable out-of-tree

stefwalter:
I'd like to run the ostree-boot-image task from a different
directory than the root of my git checkout. I'd like it
not to polute my checkout with its output and temporary
files.

This simple change should accomplish that.


arilivigni:
[test]

PR #390, opened 2017-09-28
add parallel building of stage jobs for f26,f27

scoheb:


scoheb:
[test]

scoheb:
[test]

scoheb:
[test]

scoheb:
@johnbieren note that I had to revert to hardcoding the parallel stages instead of building them in a map. Seems to be a limitation of declarative right now.
In addition, we need https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/201 released to be able to use failFast on our parallel stages...until then, even if a stage job for f27 fails, we have to wait until f26 completes before the entire build will fail...

scoheb:
[test]

scoheb:
[test]

scoheb:
closed via #414

PR #395, opened 2017-09-29
Create an image every time

johnbieren:
This forces image-compose to run every time, but it will only rsync if the conditions from before are met.  It then copies the qcow2 into the workspace.  This sets us up to allow package testing with that as the subject if it works.


johnbieren:
[test]

johnbieren:
[merge]

PR #396, opened 2017-10-02
Echo original_spec_nvr earlier

johnbieren:


pypingou:
👍 for me, thanks :)


(If we had kept the ticket open we could have close it w/ this commit (and thus link both also on GH's UI)).

arilivigni:
@pypingou added it you don't have to reopen to comment and this just decreases the window still not 100% guaranteed

johnbieren:
[merge]

pypingou:
> this just decreases the window still not 100% guaranteed

Indeed, and it's where the re-trigger mechanism becomes crucial :)

PR #401, opened 2017-10-02
Image compose needs to happen before ostree boot sanity

johnbieren:
I previously put ostree boot sanity before image compose, because I did not see why you would build an image for an ostree that doesn't even boot, but ostree boot requires an image to exist to work.


johnbieren:
[test]

johnbieren:
[test]

johnbieren:
[test]

PR #406, opened 2017-10-03
Pull fedora-atomic.ks from upstream fedora

p3ck:


p3ck:
@johnbieren it pulls it every run so we shouldn't need to remove any old versions.  But I think I'll need some other logic in here for rawhide since we need to pull from master.

johnbieren:
@p3ck The rawhide -> master connection is already there
https://pagure.io/fedora-kickstarts/blob/rawhide/f/fedora-atomic.ks

p3ck:
[merge] forgot the workflow changed

PR #419, opened 2017-10-10
Don't load job.props from duffy allocate job.

p3ck:
It only contains RSYNC_PASSWORD which overwrites the RSYNC_PASSWORD we use for artifacts server.


p3ck:
[test] not sure what caused the failure.  trying again.

p3ck:
DUFFY_HOST is in job.props unfortunately which is needed.  other than patching cciskel upstream it looks like jbieren's workaround is good enough.

PR #421, opened 2017-10-10
Add ExecutionId

scoheb:
- needed for pod name and label
- should be used for messaging to segregate messages from branches


scoheb:
[merge]

PR #426, opened 2017-10-12
Hotfix disable integration tests

johnbieren:
The integration tests in a container passed just fine in stage, but we have been getting hanging processes for the last day whenever it runs. Until we can fix that, we should disable these.


johnbieren:
@scoheb @arilivigni if you guys agree, I would like to merge this even before the stage job agrees with this change as it is affecting production.

arilivigni:
Go for it just let Dominik know  please

johnbieren:
[merge] Will do

PR #429, opened 2017-10-16
Checkout rev before running package tests

johnbieren:
In the current workflow, if the commit in question changes the tests in the dist-git repo, they wouldn't be picked up as we would just be on master.


johnbieren:
[test]

johnbieren:
[merge]

PR #432, opened 2017-10-16
Change mock's fs path to use increased pod fs

johnbieren:
We need mock to use the increased OpenShift filesystem so it can build the kernel. If this passes stage, I will manually trigger it to build a kernel in stage as well before merging to ensure this works.


config/Dockerfiles/rpmbuild/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN for i in {1..5} ; do dnf -y install ansible \
 RUN echo "config_opts['package_manager'] = 'dnf'" >> /etc/mock/site-defaults.cfg
 RUN echo "config_opts['plugin_conf']['lvm_root_opts']['size'] = '16G'" >> /etc/mock/site-defaults.cfg
 RUN echo "config_opts['plugin_conf']['lvm_root_opts']['poolmetadatasize'] = '30G'" >> /etc/mock/site-defaults.cfg
-RUN echo "config_opts['basedir'] = '/home/rpmbuild/'" >> /etc/mock/site-defaults.cfg
+RUN echo "config_opts['basedir'] = '/workDir/workspace/rpmbuild/'" >> /etc/mock/site-defaults.cfg
```
scoheb: this binds us to the string workDir which could be changed at any time....can we not make use of the logic to determine the current dir as being the workspace?

johnbieren: I suppose I could just take this line out of the Dockerfile and in the .sh file write to the defaults file and have it build in $pwd

PR #433, opened 2017-10-17
Move ostree-boot-image to launch as a pod in openshift

p3ck:
Update to use correct tag for ostree boot image
call the pod by it's correct name.
rsync the images dir from artifacts
set RSYNC_PASSWORD because bad duffy
fix up logic in boot sanity container
also fix Jenkinsfile to rsync images back to artifacts
Include /sys/class/net volume to satisfy bridge requirements


JenkinsfileStageTrigger
```diff
@@ -139,7 +151,8 @@ pipeline {
                                     string(name: 'RPMBUILD_TAG', value: tagMap['rpmbuild']),
                                     string(name: 'RSYNC_TAG', value: tagMap['rsync']),
                                     string(name: 'OSTREE_COMPOSE_TAG', value: tagMap['ostree-compose']),
-                                    string(name: 'PACKAGE_TEST_TAG', value: tagMap['package-test'])
+                                    string(name: 'PACKAGE_TEST_TAG', value: tagMap['package-test']),
```
johnbieren: You are going to have to rebase at some point. package-test is deprecated, it is all singlehost-test, so we wouldn't want that line being added back in

PR #434, opened 2017-10-17
Pull fedora-atomic-testing.json from upstream

robnester-rh:
Updated the ostree-compose.sh file to pull the
fedora-atomic-testing.json file from the upstream fedora source so that
we always have up to date data.

### DIFF
--- a/config/Dockerfiles/ostree_compose/ostree-compose.sh
+++ b/config/Dockerfiles/ostree_compose/ostree-compose.sh
@@ -61,7 +61,7 @@ gpgcheck=0
 skip_if_unavailable=False
 EOF
 
-# Get our latest fedora-atomic-testing.json file and write it to /home/output/logs/
+# Get our latest fedora-atomic-testing.json file and write it to $base_dir/logs/
 curl -o $base_dir/logs/fedora-atomic-host.json https://pagure.io/fedora-atomic/raw/${branch}/f/fedora-atomic-host.json
 
 cat << EOF > $base_dir/config/ostree/fedora-atomic-testing.json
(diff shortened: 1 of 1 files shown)


dperpeet:
Will the static list be removed as a followup? Or is that a separate thing?

https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/config/package_list

johnbieren:
Our trigger depends on the package_list file @dperpeet referenced, so there will be more work to fix the trigger than just to simply delete that file from the repo once this pr is merged

robnester-rh:
@johnbieren - Yes, @p3ck and I discussed the need to update the trigger job and there are a few options. It will be handled under another PR prior to a PR to remove the package list from this repo.

config/Dockerfiles/ostree_compose/ostree-compose.sh
```diff
@@ -61,9 +61,12 @@ gpgcheck=0
 skip_if_unavailable=False
 EOF
 
+# Get our latest fedora-atomic-testing.json file and write it to /home/output/logs/
+curl -o /home/output/logs/fedora-atomic-host.json https://pagure.io/fedora-atomic/raw/${branch}/f/fedora-atomic-host.json
```
johnbieren: $base_dir/logs not /home/output I believe

robnester-rh: @johnbieren: I had mimicked the behavior in ostree-image-compoose. I'll update and push again.

johnbieren: ostree-image-compose isn't containerized until my PR is merged, this stage already is, hence the discrepancy

robnester-rh:
[merge]

PR #444, opened 2017-10-19
Specify ostree-image-compose default libvirt bridge

johnbieren:
We should leave virbr0 open for anything that needs it; I don't believe any of our workflow should take that bridge every time.


johnbieren:
[merge]

johnbieren:
@samvarankashyap FYI. Multiple containers in the same pod were trying to use virbr0, giving Bill and me errors.  I'd recommend changing the default libvirt bridge in your PR if you are using libvirt to provision instances, leaving virbr0 open

PR #449, opened 2017-10-20
Remove some duffy things from Jenkinsfile

johnbieren:
Some initial post duffy life cleanup. I left all the functions in the src/ groovy file even though we won't be calling them.  I will raise another PR later to clean up all the paths and be consistent as we talked about


johnbieren:
[test]

johnbieren:
[merge]

PR #451, opened 2017-10-20
Specify virbr2 for bridge and don't install libvirt-daemon-config-net…

p3ck:
…work

which has the default config of virbr0

### DIFF
--- a/config/Dockerfiles/ostree-boot-image/default.xml
+++ b/config/Dockerfiles/ostree-boot-image/default.xml
@@ -1,10 +1,11 @@
 <network>
   <name>default</name>
+  <uuid>03f5e3d1-164e-4059-9df8-8c1494b81025</uuid>
   <bridge name="virbr1"/>
   <forward/>
-  <ip address="192.168.122.1" netmask="255.255.255.0">
+  <ip address="192.168.123.1" netmask="255.255.255.0">
     <dhcp>
-      <range start="192.168.122.2" end="192.168.122.254"/>
+      <range start="192.168.123.2" end="192.168.123.254"/>
     </dhcp>
   </ip>
 </network>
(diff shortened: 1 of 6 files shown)


p3ck:
[test]

p3ck:
[test]

p3ck:
[test]

config/Dockerfiles/ostree-boot-image/default.xml
```diff
@@ -1,5 +1,6 @@
 <network>
   <name>default</name>
+  <uuid>03f5e3d1-164e-4059-9df8-8c1494b81025</uuid>
```
johnbieren: do these uuids ever have to change?

p3ck: no - and they could probably even be the same between containers.

samvarankashyap:
@p3ck @johnbieren : Is there any way we could spin up the containers when we need and delete them when it finishes it job instead of running them simultaneously. Which might solve the our issue. Further it uses the less resources at any given point of time . ( I don't know whether it's applicable or not in our use case . Just a thought )

johnbieren:
@samvarankashyap my understanding was that the whole benefit of using OpenShift is that the containers are always up and available so you don't have to waste time bringing them up before using them. I am far from an SME on that front, though.

samvarankashyap:
@johnbieren : I completely agree with you . However, I also observed that now that we are building the images and storing them in openshift registry . It hardly takes seconds to boot up the containers from the image in openshift . If the images are pulled down from external registries like docker , it would take time though.

p3ck:
[test] disk actiity timeout

p3ck:
[merge]

p3ck:
[merge]

PR #453, opened 2017-10-23
Run master package tests

johnbieren:
Pretty much all tests that are ported to dist-git are only being put into the master branch. Thus, the f26 and f27 pipelines will almost never run any tests.  Proposing we run the master tests for now, just so tests are being run more often.  The only real negative to this is that we are not verifying the addition to tests to the dist git repo as a change like we would like and if there is a test that runs on master, but not f26, we will be running it anyways. I don't really think that the dist git test addition is ready yet for those scenarios, so this is likely fine for now

### DIFF
--- a/config/Dockerfiles/singlehost-test/package-test.sh
+++ b/config/Dockerfiles/singlehost-test/package-test.sh
@@ -44,8 +44,10 @@ fi
 pushd ${package}
 
 # Check out the appropriate branch and rev
-git checkout ${branch}
-git checkout ${rev}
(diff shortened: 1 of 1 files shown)


johnbieren:
@arilivigni thoughts on this?

johnbieren:
[test] wait_for port 22 timed out in boot sanity

arilivigni:
@johnbieren makes sense to me to do this in the short-term

johnbieren:
[merge]

PR #456, opened 2017-10-24
If provisioning fails then take a screen shot of the failed system.

p3ck:
also removed the $? checking since we have set +e and it won't be executed

### DIFF
--- a/config/Dockerfiles/ostree-boot-image/ostree-boot-image.sh
+++ b/config/Dockerfiles/ostree-boot-image/ostree-boot-image.sh
@@ -2,14 +2,14 @@
 
 set -xeuo pipefail
 
-brctl show
-
-base_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+#base_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+base_dir="$( pwd )"
 
 function clean_up {
     set +e
     ara generate junit - > /home/output/logs/ansible_xunit.xml
     if [ -e "$base_dir/hosts" ]; then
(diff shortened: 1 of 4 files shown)


config/Dockerfiles/ostree-boot-image/ostree-boot-image.sh
```diff
@@ -2,14 +2,14 @@
...
-
-base_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+#base_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+base_dir="$( pwd )"
 
 function clean_up {
     set +e
     ara generate junit - > /home/output/logs/ansible_xunit.xml
     if [ -e "$base_dir/hosts" ]; then
+        virsh screenshot --file /home/output/logs/atomic-host-fedoraah.ppm atomic-host-fedoraah
```
scoheb: any reason for this name: fedoraah ?

p3ck: not really.  Originally I was doing this as a block: rescue: piece in ansible and that would be the name if I was using {{ item.key }}.  I can make it just atomic-host.ppm, would that be better?

scoheb: not a big deal...atomic-host.ppm is fine.

p3ck:
[test]

p3ck:
[test] rsync file disappeared

p3ck:
[test] wake up bot

p3ck:
[merge]

PR #457, opened 2017-10-24
Collect container logs at end of build

scoheb:


scoheb:
[test] port 22 timeout

scoheb:
Yes. This is meant for troubleshooting...it applies to the jnlp container and the containers that run something other that 'cat'

scoheb:
[merge]

PR #458, opened 2017-10-25
Add IRC notifications for stage-trigger builds

scoheb:


JenkinsfileStageTrigger
```diff
@@ -30,6 +30,10 @@ tagMap['ostree-boot-image'] = STABLE_LABEL
 // Fedora Fedmsg Message Provider for stage
 MSG_PROVIDER = "fedora-fedmsg-stage"
 
+// IRC properties
+IRC_NICK = "stage-trigger-bot"
```
arilivigni: Don't we want to use the bot name we have configured in the global config.  I think it is contra-bot

robnester-rh: If this is using the same bot, agreed. However if we're adding another bot just for stage trigger notices, this should be ok, right?

scoheb: this is not using the global config since we are not using the plugin but actually poking IRC manually.
To be able to use contra-bot, then we would have to remove the global config and see when and if the contra-bot account leaves the room.

arilivigni: Not sure I follow?  Why not use the global config and then you update job status through the bot.  I was going to add it that way.  Or are you saying we don't need a global config at all anymore?

scoheb: Yes. This solution does not require the IRC plugin nor its global config. It uses straight IRC commands piped to the IRC directly. Using the post{} step of declarative we ensure it is always sent.
So, we can use the name 'contra-bot'...but before we use it, we need to make sure the plugin is not connected with it (or else there will be a conflict)

arilivigni: 1. Then let's remove the global config and the plugins if we are not using them
2. I am not sure this will work is your nick registered otherwise you may not be able to join channels especially fedora based ones

scoheb: - Config and plugins removed.
- So far, I have been able to post messages to contra-ci-cd. Have not tried others.

scoheb: The problem is that since this solution does not maintain a persistent connection, if multiple jobs attempt to post simultaneously to the channel, I "think" only one will succeed (the other fails silently). 
For this reason, that is why I had the nickname as "stage-trigger-bot". But I may have to even append the #build number to the nickname once we got parallel.
Does the actual bot name matter? We just want the notification...

As for fedora, I can test this on the side to see if we can post without a registered nick.

scoheb: was able to post to '#fedora-ci'...

scoheb:
[test] pagure 500 error

arilivigni:
Sounds like you will have to append the build number anyway then so might as will have a general bot name that is for our team?

scoheb:
or 'contra-bot-<random#>' ?

arilivigni:
contra-bot-<build#> ?

scoheb:
the problem I see is that if we expand to using this for prod-f26, prod-f27 and stage, then we "could" have a collision if any of those jobs have builds running with same number...I guess unlikely.

arilivigni:
contra-bot-<pipeline-name>-<build#> ?

scoheb:
failed test is not relevant. stage trigger job is now posting to IRC.

scoheb:
[merge]

scoheb:
Example:

(05:13:00 PM) contra-bot-5c72a: (notice) ci-stage-pipeline-trigger (PR #458 scoheb) build #645: FAILURE: https://jenkins-continuous-infra.apps.ci.centos.org/job/ci-stage-pipeline-trigger/645/

PR #461, opened 2017-10-26
Cleanup executeInContainer method

scoheb:


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -632,22 +632,19 @@ def executeInContainer(String stageName, String containerName, String script) {
...
     //
     containerEnv = env.getEnvironment().collect { key, value -> return key+'='+value }
-    sh "mkdir -p " + stageName
+    sh "mkdir -p ${stageName}"
     try {
         withEnv(containerEnv) {
             container(containerName) {
-                sh 'pwd'
-                sh 'ls -l /tmp'
                 sh "cp -fv ${WORKSPACE}/fedora.keytab /home/fedora.keytab"
```
johnbieren: Could probably do an if stageName = rpmbuild then cp, but not sure that makes it cleaner =)

scoheb: we sure it is only needed by the rpmbuild container?

johnbieren: I am pretty sure it is only ever used in the build submission to koji, which is in the rpmbuild stage. The integration and functional tests don't use it. @p3ck none of the compose or booty sanity stages use it, correct?

p3ck: correct, nothing but rpmbuild uses the fedora keytab.

scoheb:
[merge]

PR #462, opened 2017-10-27
Update ostree-compose.sh with fedora-atomic-host-base

johnbieren:
If you look at the upstream files, they now just call include on -base, so we need to get that file too. It exists on both f26 and f27.


johnbieren:
[merge]

PR #470, opened 2017-11-01
Remove diagram for Fedora pieces since this is out of date

arilivigni:
- No need to include Fedora infrastructure in the document
  this diagram is won't be kept out of date and isn't accurate


arilivigni:
[merge]

PR #473, opened 2017-11-07
limit depth to top level files and dirs.

p3ck:


p3ck:
[merge] fail has nothing to do with my change and I verified my change works.

PR #477, opened 2017-11-08
Always push a .qcow2 image to artifacts

johnbieren:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -284,14 +284,9 @@ podTemplate(name: podName,
                     stage(currentStage) {
                         // Set stage specific vars
                         pipelineUtils.setStageEnvVars(currentStage)
-                        // We always run, but don't always push to artifacts
-                        env.PUSH_IMAGE = "false"
 
                         // Check if a new ostree image compose is needed
                         if (fileExists("${env.WORKSPACE}/NeedNewImage.txt") || ("${env.GENERATE_IMAGE}" == "true")) {
-                            // We will push a new qcow2 to artifacts
-                            env.PUSH_IMAGE = "true"
-
                             // Set our message topic, properties, and content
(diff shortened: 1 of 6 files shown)


p3ck:
@arilivigni - this could have a huge impact on the storage we use on artifacts.ci.centos.org.  Has anyone calculated how much space this could end up using?  And has anyone confirmed this will be ok with CentOS?

Jenkinsfile
```diff
@@ -482,7 +471,7 @@ podTemplate(name: podName,
                     pipelineUtils.getContainerLogsFromPod(OPENSHIFT_NAMESPACE, env.NODE_NAME)
 
                     // Archive our artifacts
-                    step([$class: 'ArtifactArchiver', allowEmptyArchive: true, artifacts: '**/logs/**,*.txt,*.groovy,**/job.*,**/*.groovy,**/inventory.*', excludes: '**/job.props,**/job.props.groovy,**/*.example', fingerprint: true])
+                    step([$class: 'ArtifactArchiver', allowEmptyArchive: true, artifacts: '**/logs/**,*.txt,*.groovy,**/job.*,**/*.groovy,**/inventory.*', excludes: '**/job.props,**/job.props.groovy,**/*.example,**/*.qcow2', fingerprint: true])
```
scoheb: Are we sure we want to keep qcow files as Jenkins artifacts?

johnbieren: I thought adding it to excludes: would make it so it didn't save as an artifact, not the opposite

scoheb: Oops. My bad. You are right. Carry on!

arilivigni:
@p3ck and @johnbieren as we spoke this morning let's wait on this until we have block storage

PR #478, opened 2017-11-09
Ensure container templates present before builds

scoheb:
- create-containers.sh no longer fails if resource already exists
- JenkinsfileStageTrigger now calls create-containers.sh


arilivigni:
If I am writing the playbooks why do we need this?  Can't we just call a roll to create containers?

scoheb:
[test]

scoheb:
I misunderstood our IRC chat...I thought you were going to be calling this script from a playbook...

arilivigni:
wrote back on IRC.  Sorry I thought it was clear of course you have a lot more logic in the shell script now.  Is all this necessary?

scoheb:
[merge]

PR #483, opened 2017-11-15
Add container to build cloud images

johnbieren:
This adds a container to build rawhide cloud images with an rpm subject.  It takes a $package and $rpm_repo url location, and it builds a rawhide host with that rpm included at build time. It currently only supports including one rpm at a time, and it only supports building rawhide. It pulls from rawhide every time, nothing is cached, so it is always new and shiny. This container will be part of a pipeline that I will be creating soon. Also, to the best of my knowledge, it relies on the nvr of the rpm you are adding being > the rpm nvr from rawhide. This could likely be fixed later, if the need arises. 

@scoheb if I am missing any key openshift files for the container, please let me know.


scoheb:
There is no `create-containers.sh` to actually create the objects in Openshift.

johnbieren:
@scoheb I am going to hold off on creating that file until my PR with the actual pipeline in it, in case I have to add more containers. I will note that down as a requirement, though. Thanks

PR #484, opened 2017-11-16
Add ability to use python3 for package tests

johnbieren:
This is mostly in preparation of the systemd pipeline. If the test subject image has no python, only python3, we will need the ability to pass this in to the playbook command.  This is the case with rawhide images. With this change, we could just define a variable python3=yes pre container call and everything should work.


johnbieren:
[merge] failures are e2e testing related

PR #492, opened 2017-11-27
Add ability to set minishift base device size

scoheb:
This will let users configure how large the rootfs needs to
be for pods in minishift


scoheb:
@arilivigni  Yeah, do not think that is necessary. Thanks

scoheb:
[merge]

PR #494, opened 2017-11-27
Protect against null object errors

scoheb:
If setStageEnvVars() is the last item in a stage section
and there is no submap for the stage, then Jenkins will throw
a cryptic 'java.io.NotSerializableException: org.codehaus.groovy.runtime.NullObject' error

This is due to the unchecked .get() that otherwise is silently ignored
if another step comes after setStageEnvVars() in a stage.


scoheb:
[merge]

PR #495, opened 2017-12-04
config/s2i/jenkins: rename jenkins bc templates

thrix:
The patch for adding dev_setup forgot to rename these template files.


scoheb:
you need to actually change `config/s2i/jenkins/continuous-infra-create-openshift-project` since `dev-setup` relies on the current filenames.

thrix:
ah, I had to be blind, fixed

PR #496, opened 2017-12-05
Add necessary changes to add systemd pipeline

johnbieren:
All of these changes are necessary in order to add support for the systemd pipeline.  However, I did not include any of the systemd Jenkinsfile type files, I will in a follow up PR.

Things done in this PR:
- Add sym link to cloud-image-compose script so we have a reliable file to call in the Jenkinsfile
- Add the inquirer container. This container has two scripts: one polls a url, waiting for it to exist, exiting if it does not by $timeout; the second gets the NVR of an rpm using repoquery from a $rpm_repo
- ostree-boot-image changes: Only run the commit verify playbook if $commit is defined; if $expected and $package are both defined, run the rpm-verify playbook (new) that will check that the running nvr of the $package rpm on the host is $expected
- Add a modified beakerlib role main.yml file to the singlehost-test container. This is only used in the upstreamfirst-test.sh script by the container. The point of this is that it hardcodes the location of restraint repo, as rawhide is at time of writing f28, and copr can't build f28 yet. Other than that it is the same as master
- Update standard-inventory-qcow2 in the singlehost-test container. The current one we are storing does not work on cloud hosts. The current one in master from pagure requires tty, so this is a modified version of master which works on both cloud hosts and with no tty
- Add upstreamfirst-test.sh to the singlehost-test container. This script gets the tests from upstreamfirst.fedorainfracloud.org instead of dist-git, as the systemd ask was to pull the tests from there.

Nothing here should break production, but I still would like to see it pass stage just to be doubly sure. 

### DIFF
--- a/config/Dockerfiles/cloud-image-compose/cloud-image-compose.sh
+++ b/config/Dockerfiles/cloud-image-compose/cloud-image-compose.sh
@@ -32,8 +32,13 @@ virtlogd &
 
 chmod 666 /dev/kvm
 
+imgname="fedora-cloud-rawhide-$package"
+
 function clean_up {
   set +e
+  pushd ${base_dir}/images
+  ln -sf $imgname.qcow2 untested-cloud.qcow2
+  popd
   kill $(jobs -p)
   for screenshot in /var/lib/oz/screenshots/*.ppm; do
       [ -e "$screenshot" ] && cp $screenshot ${base_dir}/logs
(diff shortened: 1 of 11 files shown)


scoheb:
LGTM

johnbieren:
[merge] it made it far enough in the stage job to valid that my changes don't break anything. Hitting infra failures left and right this week, but no errors due to this

PR #499, opened 2017-12-11
Exit poller nonzero if timeout hit

johnbieren:
Lost that part of my script with a rework. If the remote_file is never found, we need to exit nonzero once we hit the timeout, as that is kind of the entire point.


johnbieren:
[merge] failures not related to this change

PR #500, opened 2017-12-11
Add all reqs for standard-test-roles Dockerfile

johnbieren:
I somehow didn't have all the official requirements listed in the Dockerfile install and never hit a bug because they have just always been there luckily, so fixing it to be cleaner.


johnbieren:
[merge]

PR #502, opened 2017-12-15
Turn off two tests for systemd that fail on rawhide

johnbieren:
See BZs from comments, these always fail on rawhide for now


johnbieren:
[merge]

PR #504, opened 2017-12-18
Add dnf-plugins-core to container

johnbieren:
dnf copr is no longer available with the base install of dnf it seems, so add dnf-plugins-core to get it.


johnbieren:
[merge] failures unrelated

PR #507, opened 2018-01-04
Get package list for trigger from upstream

johnbieren:
This will first try to use the upstream package list json to determine if the package is part of atomic. If the checking of upstream fails, it falls back to checking our static list.

### DIFF
--- a/JenkinsfileTrigger
+++ b/JenkinsfileTrigger
@@ -83,7 +83,7 @@ node('master') {
                             fi
 
 
-                            # Save the bramch in job.properties
+                            # Save the branch in job.properties
                             echo "branch=${branch}" >> ${WORKSPACE}/job.properties
 
                             # Verify this is a branch in our list of targets defined above in the parameters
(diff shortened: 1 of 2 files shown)


JenkinsfileTrigger
```diff
@@ -93,12 +93,28 @@ node('master') {
...
+                                pushd fedora-atomic
+                                git checkout ${fed_branch}
+                                popd
+
+                                python ${PROJECT_REPO}/utils/package_checker.py ${fed_repo}
+                                CHKR_RC=$?
+                                # If $? -eq 0, we care about package
+                                if [ $CHKR_RC -eq 0 ]; then
+                                    valid=1
+                                # If $? -eq 2, upstream package list didnt exist so use legacy method to check
```
scoheb: nit: didnt -> didn't

utils/package_checker.py
```diff
@@ -0,0 +1,27 @@
...
+# Exit 1 if file exists, but package isnt in the list
+
+jsonpath = 'fedora-atomic/fedora-atomic-host-base.json'
+
+# Make sure json file exists
+if not os.path.isfile(jsonpath):
+    print("Could not find upstream package json file")
+    sys.exit(2)
+
+jsonfile = open(jsonpath, 'r').read()
```
robnester-rh: When opening/reading files in python, it's better to use the context manager approach. This handles closing a file if an error occurs, etc. This section would be better written as:

```
with open(jsonpath, 'r') as f:
    atomicjson = json.load(f)
```

utils/package_checker.py
```diff
@@ -13,15 +13,15 @@
...
-# Check both all packages and x86_64 specific packages
-if mypackage in atomicjson["packages"] or mypackage in atomicjson["packages-x86_64"]:
-    print ("Package of interest!")
-    sys.exit(0)
-
-# Fail if not
-sys.exit(1)
+with open(jsonpath, 'r') as f:
+    atomicjson = json.load(f)
+    mypackage = sys.argv[1]
```
robnester-rh: This should be moved out of the `with open `context. this should read:

```
with open(jsonpath, 'r') as f:
    atomicjson = json.load(f)

mypackage = sys.argv[1]
```

johnbieren:
[merge] errors seem to be a fedora stg datagrepper issue

PR #510, opened 2018-01-18
Fix slave FROM tag and set JVM arch for master

scoheb:
Since v3.7 of openshift, there is the possibility for breakage in
backwards compatibilty when using the v3.7 'oc' command with a 'v3.6'
master.

This change forces us to use the v3.6 oc binary

This is a better practice in any event, since we should not be pointing
to a 'latest' tag.

See https://github.com/openshift/jenkins-sync-plugin/issues/173 and https://github.com/openshift/jenkins/issues/477 for details


scoheb:
[test]

PR #511, opened 2018-02-06
Specify centos7 jenkins:2 image in s2i

johnbieren:


scoheb:
one thing to keep in mind is that the new label uses 3.7 which is fine for Upshift. The Centos instance is still using 3.6...so not sure of impact of using 3.7 with a 3.6 openshift instance...

PR #512, opened 2018-02-07
Allow insecure git checkout for Jenkinsfile

johnbieren:
This has to be done if your Jenkinsfile lives in a repo that your jenkins master cannot securely connect to. Unless there are any drawbacks to this (Scott?), I think we should make this change.


johnbieren:
[merge]

PR #515, opened 2018-02-15
Update standard-test-roles before running tests

johnbieren:
There was a failure due to standard-test-roles being too old in the container, so I was asked to update standard-test-roles to latest each time. Kind of goes against the rules of how OpenShift should work in my opinion, so instead of stripping standard-test-roles from the Dockerfile and doing dnf -y install standard-test-roles in the shell script, I decided to do an update in the script and rpm -q so they can see the version in case there is no update available.


johnbieren:
[merge]

PR #516, opened 2018-02-15
pipelineUtils: automatically figure out namespace and internal registry URL

psiroky:
This is an initial version, intended to get some feedback. I have _not_ yet tested this locally. Will setup my local env and do that next.

cc @scoheb

### DIFF
--- a/src/org/centos/pipeline/PipelineUtils.groovy
+++ b/src/org/centos/pipeline/PipelineUtils.groovy
@@ -1071,3 +1071,44 @@ def checkTests(String mypackage) {
     return sh (returnStatus: true, script: """
     git ls-remote --exit-code -h https://upstreamfirst.fedorainfracloud.org/${mypackage}""") == 0
 }
+
+/**
+ * Figures out current OpenShift project name. That is the project the Jenkins is running in.
+ * The method assumes that the Jenkins instance has kubernetes plugin installed and properly configured.
+ */
+def getDefaultOpenShiftProject() {
+    return openshift.withCluster() {
+        openshift.project()
+    }
+}
+
+/**
+ * Figures out the internal Docker registry URL which is supposed to host all the images for current OpenShift project.
+ *
+ * The method tries few default hostnames which are known to work on different OpenShift versions and if none of those
+ * is resolvable, it falls back to a "hacky" way, based on the content of ImageStream. The fallback approach assumes
+ * that all images in the current project are stored in the internal Docker registry (as it just looks at the first
+ * imagestream it finds). This is not 100% bullet proof, but should be good enough as a fallback.
+ */
+def getOpenShiftDockerRegistryUrl() {
+    def knownHostnames = ["docker-registry.default.svc.cluster.local", "docker-registry-default"]
+    for (String hostname : knownHostnames) {
+        node("master") { // the shell script below needs to run on an actual node
+            def status = sh(script: "getent hosts $hostname > /dev/null", returnStatus: true)
+            if (status == 0) {
+                return "$hostname:5000"
+            }
+        }
+    }
+    // fallback to the "hacky" way, based on content of ImageStream
+    return openshift.withCluster() {
+        def someImageUrl = openshift.raw("get imagestream -o=jsonpath='{.items[0].status.dockerImageRepository}'").out
+        String[] urlParts = someImageUrl.split('/')
+        if (urlParts.length != 3) { // there should be three parts in the image url: <docker-registry-url>/<namespace>/<image-name:tag>
+            throw new IllegalStateException("Can not determine Docker registry URL! Unexpected image URL: $someImageUrl " +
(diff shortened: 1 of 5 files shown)


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1060,3 +1060,34 @@ def watchForMessages(String msg_provider, String message) {
...
+/**
+ * Figures out the Docker registry URL which is supposed to host all the images for current OpenShift project.
+ *
+ * The method uses the 'jenkins' image and assumes that it is part of the current project and that other images are
+ * in the same project as the 'jenkins' image. This is not 100% bullet proof, but should be good enough as starting point.
+ *
+ */
+def getOpenShiftDockerRegistryUrl() {
+    return openshift.withCluster() {
+        def jenkinsImageUrl = openshift.raw("get is jenkins --template='{{.status.dockerImageRepository}}'").out
```
psiroky: This is way too fragile, as different projects can have different images names. I'll try to figure out  better way to get the image registry.

psiroky: Slightly different solution (not relying on the presence of any specific container):
1) Create dummy imagesteam: `oc create imagestream dummy`
2) Get the registry URL from the `dummy` imagestream: `oc get imagestream dummy -o=jsonpath="{.status.dockerImageRepository}"` + parse the URL (same as above)
3) Remove the imagestream `dummy`.

This is definitely less error prone, but it requires two additional API calls every time. I am not sure if this can be an issue or not though. Any comments more than welcome.

johnbieren:
[test]

johnbieren:
[test]

src/org/centos/pipeline/PipelineUtils.groovy
scoheb: I am not crazy about creating/removing a dummy imagestream.
I did find this: https://github.com/openshift/jenkins/commit/68d730e67597970f790d5660257869ceb1f67d37#diff-8527116e66492691f95b5090ad6d719e

- you can get all the names of imagestreams.
- you loop over each name until you find what you want based on your split method check to record your docker reg URL.
- once you find one, you exit your loop.

psiroky: I am not sure I understand -- how do I decide which imagestream name is the correct one? The URL will likely have the same structure, being it the internal docker registry, or any other (external) registry. Without knowing e.g. which imagestream I should look at, I don't see way to get the correct URL.

scoheb: I guess I have never seen an Openshift tenant that had imagestreams that were not local...

psiroky: We were using that extensively in middleware. The images were built elsewhere and our imagestreams pointed to external registry. But that's of course not the case here, so we should be fine.

scoheb: Then we should be careful then...If we go the create/delete routem, we would need to ensure the dummy is a unique name...

psiroky: Don't get me wrong -- I don't like the create/delete route :)

After thinking about this more, I would actually just leave it as is for now as it works. If we encounter issues in future then we can perhaps fallback to the create/delete option.

psiroky:
I believe the build failure is not related to changes in this PR.

johnbieren:
Agreed, LGTM if it looks good to @scoheb

psiroky:
@arilivigni @scoheb WDYT? Is this good enough to merge, or do you want me to do some additional tests/changes?

psiroky:
@scoheb I've tried to use the default hostnames as we talked about, but there is another issue -- I could not find a sandbox-friendly way to check if the hostname is valid. The obvious `InetAddress.getByName(hostname)` fails with `org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod java.net.InetAddress getByName java.lang.String`. I'll try to think of a different approach, but out of ideas for now.

johnbieren:
@psiroky We should be able to just approve that groovy script to be run, in my opinion. Agree @scoheb ?

psiroky:
@johnbieren yeah, we can whitelist that particular signature. It kind of breaks the OOTB functionality as every Jenkins instance using this library/method needs to be configured -- but if we are OK with that, then I can just add the method as whitelisted and that should work (after re-deploying the jenkins image).

scoheb:
or can we not simply use some shell command to see if name is resolved?

psiroky:
We can do that, with e.g. `getent hosts $hostname`. However, IIUC that brings its own downsides as the shell script needs to actually run on a specific node (inside the `node()`). So we would either need to hardcode it into the method itself (I haven't tried, but I think it should be possible) -- or we would need to call the methods from `node()` block inside the `Jenkinsfile`. I don't particularly like either of these options. I'll try the first one though, to at least figure out if it indeed is a working alternative.

PR #518, opened 2018-02-20
Adding ability to send metrics to influxdb

joejstuart:


vars/ciMetrics.groovy
```diff
@@ -0,0 +1,51 @@
...
+} finally {
+    ciMetrics.writeToInflux()
+}
+ */
+class ciMetrics {
+
+    // A map to store the data sent to influx
+    def customDataMap = ["ci_pipeline": [:]]
+    // This will prefix the data sent to influx. Usually set to the job name.
+    def prefix = "ci_pipeline"
```
robnester-rh: I'd perhaps set this to default to the job name, and allow overiding?

vars/ciMetrics.groovy
```diff
@@ -0,0 +1,51 @@
...
+    // The influx target configured in jenkins
+    def influxTarget = "localInflux"
+    // The influx database that will be written to
+    def measurement = "ci_pipeline"
```
robnester-rh: Could this be configured such that it is based on the job name as well? Or would it make more sense to leave it as a non-variable database name? I'm not sure what your influx database provisioning code looks like.

vars/ciMetrics.groovy
joejstuart: Good suggestion. I will update this.

joejstuart: I'm going to leave the defaults as is for now. Maybe just an update to the docs. Since this will prefix the job name.

PR #520, opened 2018-02-21
Make test_subjects potentially variable in upstreamfirst tests

johnbieren:
I want to be able to pass in TEST_SUBJECTS and use that instead of always relying on the hardcoded value.


johnbieren:
[merge]

PR #521, opened 2018-02-23
Add topic as parameter to sendMessage

johnbieren:
Currently, we just get lucky and depend on another function in the library setting topic=


johnbieren:
[merge]

PR #522, opened 2018-02-23
Change upstreamfirst error codes

johnbieren:
I want to be able to leverage the error code of 2 if the ansible-playbook fails, so I need *only* the ansible-playbook call to result in the script returning 2.


johnbieren:
[merge] This script isn't in any of the ci-pipeline code so the stage job is irrelevant once the container builds (which it did)

PR #525, opened 2018-03-01
Add util to retrieve github integration status

cevich:
This can be used (for example) to gate further testing upon success of
integrated tasks, such as travis CI results or CLA checking.

Signed-off-by: Chris Evich <cevich@redhat.com>


cevich:
I find it handy, maybe useful here?  Maybe not?  No worries either way.

PR #531, opened 2018-03-09
Metrics stack containers/s2i templates

joejstuart:
This is still WIP. There's three containers. Influxdb/Grafana/Telegraf. The Telegraf container is just to collect jenkins system metrics at the moment.


config/s2i/influxdb/continuous-infra-create-openshift-project
```diff
@@ -0,0 +1,21 @@
+PROJECT=continuous-infra
```
johnbieren: I don't think this file will work in its proposed place because there are no jenkins templates at its level and it looks like a dupe of the one in config/s2i.. Is this file meant to be here?

joejstuart: Good catch. This shouldn't be there.

PR #532, opened 2018-03-09
Add fed_ variables for koji task, update checkTests function

johnbieren:

### DIFF
--- a/src/org/centos/pipeline/PipelineUtils.groovy
+++ b/src/org/centos/pipeline/PipelineUtils.groovy
@@ -1072,5 +1072,5 @@ def watchForMessages(String msg_provider, String message) {
 def checkTests(String mypackage, String mybranch, String tag) {
     echo "Currently checking if package tests exist"
     return sh (returnStatus: true, script: """
-    wget -O- -q https://src.fedoraproject.org/rpms/${mypackage}/raw/${mybranch}/f/tests/tests.yml | grep "\- ${tag}"""") == 0
+    wget -O- -q https://src.fedoraproject.org/rpms/${mypackage}/raw/${mybranch}/f/tests/tests.yml | grep "\- ${tag}" """) == 0
 }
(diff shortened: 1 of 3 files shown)


johnbieren:
@robnester-rh having the 4 "s in a row in the script sh thing should be fine, right? I think it is syntactically correct..

src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1062,12 +1062,15 @@ def watchForMessages(String msg_provider, String message) {
...
+ * @param tag
  * @return
  */
-def checkTests(String mypackage) {
-    echo "Currently checking if an upstreamfirst repo exists"
+def checkTests(String mypackage, String mybranch, String tag) {
+    echo "Currently checking if package tests exist"
     return sh (returnStatus: true, script: """
-    git ls-remote --exit-code -h https://upstreamfirst.fedorainfracloud.org/${mypackage}""") == 0
+    wget -O- -q https://src.fedoraproject.org/rpms/${mypackage}/raw/${mybranch}/f/tests/tests.yml | grep "\- ${tag}"""") == 0
```
joejstuart: Might need single quotes around wget since double quotes are being used in grep.

robnester-rh: If you're using the """
"""

approach, then the code contained within should be formatted in whatever way it would be devoid of the """

joejstuart: :+1:

robnester-rh: Actually, I shouldn't have just glanced. I think this is syntactically incorrect as the """ will terminate on the first three """ encountered. I think that you might want to amend this line to:

```wget -O- -q https://src.fedoraproject.org/rpms/${mypackage}/raw/${mybranch}/f/tests/tests.yml | grep "\- ${tag}" """) == 0```

johnbieren: I'm not sure the ${tag} will expand if I use single =|

joejstuart: Yeah. It doesn't look like it will.

johnbieren:
[merge]

PR #537, opened 2018-03-14
Provide nvr and original_spec_nvr in new build scripts

johnbieren:


johnbieren:
Because we promised to provide both nvr and original_spec_nvr and I am not modifying spec files in this script so they are the same
https://github.com/CentOS-PaaS-SIG/ci-pipeline#orgcentosprodcipipelinepackagerunning
Good catch on it only being RPM_NAME in the local one. I will fix that

johnbieren:
[merge]

PR #539, opened 2018-03-16
Add skip stage function

johnbieren:
In scripted pipeline, you have no way to mark a stage as skipped like you do in declarative. To my google searching, it simply isn't supported. But, with this function, you can. To do it, you would do something like this
stage('foo') {
    if (skipme) {
        pipelineUtils.skip('foo')
    }
}

@robnester-rh Do I need the import in the vars file too or just the src/..... .groovy file?

Also note that this requires a groovy script approval (for those of you using s2i). It is staticMethod org.jenkinsci.plugins.pipeline.modeldefinition.Utils markStageSkippedForConditional java.lang.String


johnbieren:
[merge]

PR #540, opened 2018-03-16
Legitimately skip functional test stage if tests DNE

johnbieren:
Stage will be present but skipped now


johnbieren:
[merge]

PR #541, opened 2018-03-16
Add tags property

joejstuart:
This will allow a pipeline to submit custom tags to influxdb


joejstuart:
This PR will allow for custom tags to be sent in with each pipeline.

PR #543, opened 2018-03-19
Need package defined for rsyncing artifacts

johnbieren:
Not sure how we got this far without this being defined honestly


johnbieren:
[merge]

PR #544, opened 2018-03-19
Enable not rawhide building

johnbieren:
Overlooked the fact that I never enabled non rawhide image building in this container


johnbieren:
Nothing in here is called in the stage job

johnbieren:
[merge]

PR #548, opened 2018-03-21
Change to f27

johnbieren:


johnbieren:
[merge] The failure had nothing to do with the rpmbuild stage which is all this change affects

PR #549, opened 2018-03-22
Make tag configurable

johnbieren:
Currently, --tags atomic is hardcoded. Make it so we can override that for upstream all packages pipeline


johnbieren:
[merge]

PR #553, opened 2018-03-23
Bash logic

johnbieren:
Not sure if it is the version of bash or what, but it is putting one dir inside the other, which is not the behavior I see locally, so be specific


johnbieren:
[merge] nothing in stage job is affected by this

PR #554, opened 2018-03-23
Need to override rpm_repo too

johnbieren:


johnbieren:
[merge] this has nothing to do with the stage job

PR #560, opened 2018-04-11
Move check package test results to pipelineUtils

joejstuart:


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1120,3 +1119,20 @@ def parseTestLog(def fileLocation) {
...
+def checkTestResults(Map testResults) {
+    def buildResult = null
+
+    testResults.each { test, result ->
+        if (result != 'PASSED') {
+            buildResult = 'UNSTABLE'
+        }
+    }
+
+    return buildResult
```
johnbieren: So then we update the actual build result with buildResult ?: something right? Just trying to make sure that if all tests pass, we aren't setting the build description to be null. I *think* we could do SUCCESS if all passed, as in theory no failures have happened up to that point, then, and some future stage could fail and turn it to failed anyways

joejstuart: yep. So if a build makes it to the end of the stages it will set buildResult to SUCCESS if it's not already set by a test failure.

johnbieren: lgtm

PR #562, opened 2018-04-16
Add function to check if fork

johnbieren:
Example
"path": "/srv/git/repositories/forks/churchyard/rpms/python34.git", 

We don't want to trigger if the git.receive message is from a fork


johnbieren:
[test] don't like the error stage job shows. I will add a PR in a little bit to change the JenkinsfileTrigger for the atomic ones to take advantage of this change as well

johnbieren:
[merge]

PR #564, opened 2018-04-16
Fix python3 call

johnbieren:
I only fixed this in package-tests.sh when I fixed this. Fix it here too


johnbieren:
[merge] jenkins is back

PR #565, opened 2018-04-17
Better package discovery

johnbieren:
Sometimes, the fed_repo isn't the package name, such as with krb5. So, I was asked to discover the rpms to install like this instead.


johnbieren:
[merge] stage job doesn't use this container

PR #566, opened 2018-04-18
Add tools needed to build rpms from pagure PRs

johnbieren:
Needed a new rpm build script as well as a function to inject the other type of fedmsg. Might be some additions later once I get more into triggering off of commits added and comments. Just going with new PRs opened to start


johnbieren:
[merge]

PR #569, opened 2018-04-19
Clean up Jenkinsfiles

johnbieren:
The only functional difference should be that in the package-tests stage, if test.log (summary output) doesn't exist, set status to UNSTABLE, if integration-tests.sh script call fails, status = UNSTABLE, and same for openshift-e2e-tests script.
Rest is just for readability


Jenkinsfile
```diff
@@ -444,7 +509,14 @@ podTemplate(name: podName,
                         pipelineUtils.sendMessageWithAudit(messageFields['topic'], messageFields['properties'], messageFields['content'], msgAuditFile, fedmsgRetryCount)
 
                         // Run integration tests
-                        pipelineUtils.executeInContainer(currentStage, "singlehost-test", "/tmp/integration-test.sh")
+                        try {
+                            pipelineUtils.executeInContainer(currentStage, "singlehost-test", "/tmp/integration-test.sh")
+                        } catch (e) {
+                            // Should improve to actually parse result files, but for now if the script failed, call it a test failure
+                            buildResult = 'UNSTABLE'
```
joejstuart: Looking at the "UNSTABLE" logic again, I think we can get rid of the buildResult variable and just set currentBuild.result. Then in the main finally currentBuild.result can be set with:
`currentBuild.result = currentBuild.result ?: 'SUCCESS'`

johnbieren:
[merge]

PR #577, opened 2018-04-24
Update singlehost test to f27

johnbieren:
There were hacks in place for booting cloud images and for the beakerlib role. STR has since fixed this, but with the older container we wouldn't get the fix.


johnbieren:
[merge]

PR #578, opened 2018-04-24
Generalize injectPRVars

johnbieren:
The less versions of this function we have to maintain the better


joejstuart:
Looks good to me. The only question I have is should we throw an exception if the message does not contain the pullrequest key?

johnbieren:
[merge]

PR #580, opened 2018-04-24
Clone and grep for checkTests

johnbieren:
This still takes ~9 seconds per run, and --list-tags doesn't seem to work, but this might suffice. It will have to be different for PRs still. Might be able to make mybranch = something for PRs to work


johnbieren:
[merge]

PR #584, opened 2018-04-27
Use koji download-build

johnbieren:
download-task doesn't work on tasks over some age, but this does


johnbieren:
[merge] stage job failed for messaging which isn't affected here

PR #588, opened 2018-04-30
checkTests bug fix

johnbieren:


johnbieren:
[test] failed in boot sanity

johnbieren:
[merge] seems to be a persistent problem in boot sanity

PR #590, opened 2018-04-30
Let checkTests work for PRs

johnbieren:
I think this way leaves the least amount of cleanup and changes to the pipelines to be compatible


johnbieren:
We need to also change package-tests.sh in singlehost test to run PR tests and change the all packages pipeline to use this

johnbieren:
[merge]

PR #594, opened 2018-05-01
Add locking functions

johnbieren:

### DIFF
--- a/src/org/centos/pipeline/PipelineUtils.groovy
+++ b/src/org/centos/pipeline/PipelineUtils.groovy
@@ -1200,6 +1200,67 @@ def skip(String stageName) {
 }
 
 /**
+ * Lock a directory on localhost
+ * @param fileLocation - The location to store the lock file
+ * @param duration - The number of seconds that if lock is this age, overwrite it
+ * @return myuuid - Generate uuid
+ */
+def obtainLock(String fileLocation, int duration) {
+    echo "Currently in obtainLock function"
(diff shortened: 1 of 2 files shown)


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1199,6 +1199,67 @@ def skip(String stageName) {
...
+                # Break if lock file is too old
+                if [ $ageDiff -ge ${duration} ]; then
+                    break
+                fi
+            done
+            # Now, either lock file is older than duration
+            # or the lock is gone, so proceed
+            echo ${myuuid} > ${fileLocation}
+            testuuid=$(cat ${fileLocation})
+            # If uuid matches, we got the lock
```
p3ck: there is still a race condition here.  It's possible something else could write to the file after you read it.  I know I gave you some of this originally.. so I feel guilty. :-)  The window is pretty small but could happen.
might want to look at using flock for a truly robust solution?

https://dmorgan.info/posts/linux-lock-files/

johnbieren: I don't think I can use fancy commands like flock since this will have to run not on a pod. Gotta create my own flock. A jock, per say

johnbieren: Maybe I can, looks like flock is a pretty common package

johnbieren:
[merge]

PR #595, opened 2018-05-02
non-capture matcher plus parsing koji-shadow builds

joejstuart:


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1333,15 +1333,15 @@ def injectArray(String prefix, def message) {
 def repoFromRequest(String request, String prefix) {
 
     try {
-        def gitMatcher = request =~ /git.+?\/([a-z0-9A-Z_\-]+)(\.git|\?).*/
-        def cliMatcher = request =~ /cli-build.+?\/([a-zA-Z0-9\-_]+)-.*/
+        def gitMatcher = request =~ /git.+?\/([a-z0-9A-Z_\-]+)(?:\.git|\?).*/
```
johnbieren: Does this make it so ++ like we saw in that one will work too? Instead of just one +

src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1333,15 +1333,15 @@ def injectArray(String prefix, def message) {
...
+        def buildMatcher = request =~ /(?:koji-shadow|cli-build).+?\/([a-zA-Z0-9\-_]+)-.*/
```
johnbieren: No idea if koji-shadow is just for the non primary koji instances, but good to have the code to support it if not. +1

src/org/centos/pipeline/PipelineUtils.groovy
joejstuart: No. I will update it for that. Thanks for the reminder.

PR #596, opened 2018-05-02
upstream-fedora-pipeline stage merge

johnbieren:
If merged, we will be able to use [upstream-merge] for PRs to cloud-image-compose (or inquirer, once we reenable that). I will add the stage trigger in another PR; wanted to get this working for us first


johnbieren:
[merge]

PR #598, opened 2018-05-02
Also get noarch rpms

johnbieren:


johnbieren:
[merge] stage job doesn't use this script

PR #601, opened 2018-05-02
Add stage trigger

johnbieren:
Kind of guessing on the ! changeset thing, but if none of the containers of interest are changed, we have no reason to run, and that has to be incorporated somehow. So, that's what I came up with


joejstuart:
There's an Included/Excluded regions textbox in the pull request builder that can be used. I've never used it, so I don't know if it would be better or not.

johnbieren:
[merge]

PR #605, opened 2018-05-04
When not syntax

johnbieren:


joejstuart:
I haven't tested it, but the docs say all child conditions must return true in the when directive for the stage to run. Sounds like the anyOf directive used with it might work.

https://jenkins.io/doc/book/pipeline/syntax/#when

johnbieren:
[merge]

PR #611, opened 2018-05-07
Use double quotes so vars resolve

johnbieren:
It seems with ''' the parameters passed in to the function don't resolve


johnbieren:
[merge] this is on openshift e2e tests and isn't called in any stage anyways. Need it merged for other work

PR #613, opened 2018-05-07
Replace JsonSlurper functions with read JSON

14rcole:


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -3,7 +3,7 @@ package org.centos.pipeline
 
 import org.centos.*
 
-import groovy.json.JsonSlurper
+import groovy.json.*
```
joejstuart: I think this import can be deleted now.

14rcole:
[test]

p3ck:
[test]

johnbieren:
[test] I aborted this since it was running with (potentially now fixed) bad rsync auth

14rcole:
@johnbieren Is that why the build keeps timing out?

johnbieren:
@14rcole as far as I am concerned, this passed atomic stage job. It timed out in the openshift e2e tests, which has been happening a bit lately, and I am not sure why. So, this is probably fine to merge. I am going to do an [upstream-test] just for sanity with the other pipeline, but should be fine.

14rcole:
@johnbieren Sounds good. @joejstuart thoughts?

PR #615, opened 2018-05-08
More escaping for lock function

johnbieren:
With these changes, the function worked for me
I also removed set -x for log size


johnbieren:
[merge]

PR #616, opened 2018-05-11
Add more s2i

johnbieren:
I did not update the plugins
I did not add something to init.groovy to make CI Pipeline the default view
And I think, but am not certain, that the init.groovy stuff for adding views is correct


johnbieren:
Have confirmed the view stuff works. Looking at fixing the other two things from my initial comment

johnbieren:
Adding some improvements and testing, will reopen after

johnbieren:
Tested in minishift. Plugins are updated

PR #618, opened 2018-05-14
adding init container. removing privileged from main container

joejstuart:
This allows the long running container to run without privileged mode set to true.


joejstuart:
[merge]

PR #619, opened 2018-05-14
Fix view s2i

johnbieren:
Also made the script approvals more minimal (too minimal to even run out of the box) because having too many seems to break things


johnbieren:
[merge]

PR #621, opened 2018-05-16
If download-build fails, use download task

johnbieren:
Have seen some weird behavior where download-build --task-id foo fails, but download-task foo works


johnbieren:
[merge]

PR #625, opened 2018-05-21
Add memsize to virt customize

johnbieren:
This may help reduce the number of memory issues we see


johnbieren:
[upstream-test]

johnbieren:
[upstream-merge]

PR #626, opened 2018-05-21
Add expected vs running bits to container

johnbieren:
Need this stuff in the container to call from the Jenkinsfiles.

Moved it all out of ostree-boot image container so there is no need to add that container to the pod just for this. Will reduce the resource requirements for OpenShift


johnbieren:
[merge]

PR #634, opened 2018-05-23
Update rpm verifier

johnbieren:
We can't use package name to verify rpms because the src rpm name doesn't always match the rpms we install. So instead, provide the location to the rpms and check that at least one rpm from that repo is installed.


johnbieren:
[merge]

PR #638, opened 2018-05-30
Revert "Replace JsonSlurper functions with read JSON"

johnbieren:
Reverts CentOS-PaaS-SIG/ci-pipeline#613

This broke the all packages pipeline. I will investigate why but I need to revert for now


johnbieren:
[merge]

PR #640, opened 2018-05-30
If status not Open, don't trigger pr pipeline

johnbieren:


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -472,7 +472,7 @@ def checkUpdatedPR(String message, String keyword) {
 
     if (ci_data['pullrequest']['comments']) {
         // Check if this comment is a merge notification
-        if (ci_data['pullrequest']['status'] == 'Merged') {
+        if (ci_data['pullrequest']['status'] != 'Open') {
```
joejstuart: Is there ever a case when ci_data['pullrequest']['comments'] == false and ci_data['pullrequest']['status'] == Open? I'm just wondering if 475 should be before 473?

johnbieren: In the if I am just checking if the key exists, meaning there is at least one comment. There is the case when there are no comments and status is open, which would be the new PR, whose trigger doesn't  call this function. But do you think I should swap it for correctness's sake? I am assuming status is always there

johnbieren:
[merge]

PR #643, opened 2018-05-30
readJSON not slurper

johnbieren:
Gotten approval for this twice, just keep having to revert and catching bugs


johnbieren:
[merge]

PR #648, opened 2018-06-05
readJSON can't parse newline

johnbieren:
The last thing to fix is jsonslurper handles something like comment: "Here is my comment\n" just fine, but readJSON does not


johnbieren:
Unfortunately message bus messages can still have \n in them. If \\n is already present in the message, it seems to work fine (to my surprise)

johnbieren:
[merge]

PR #650, opened 2018-06-06
Changed PipelineUtils.obtainLock to ensure that directory exists

dirgim:
* Changed PipelineUtils.obtainLock to execute mkdir -p before checking/creating lock


dirgim:
[merge]

PR #651, opened 2018-06-06
Add better checking in the trigger

johnbieren:
Currently, if you hear a commit to the cockpit repo for example, you check the atomic host json and say there are no packages named cockpit, so do not trigger. However, there are rpms in atomic host from the cockpit distgit repo, they just aren't named "cockpit," they are "cockpit-foo." So, query the PDC if we don't find it in the json originally to see if the package is one of these examples. This adds ~20 seconds sadly.

### DIFF
--- a/JenkinsfileTrigger
+++ b/JenkinsfileTrigger
@@ -114,7 +114,7 @@ node('master') {
                                 pushd fedora-atomic
                                 popd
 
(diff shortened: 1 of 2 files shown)


johnbieren:
Switched to using mdapi per Pierre's suggestion as it adds a lot of speed up.

johnbieren:
[merge]

PR #652, opened 2018-06-06
Add oc build recycler

johnbieren:
Once a day, this job will remove all builds in the openshift instance with -build in their name that are in the completed state. I didn't do the error state ones in case someone wants to see the log output. I also didn't put any age restrictions because I don't see the point in keeping around a successfully completed build for any time at all.

### DIFF
--- a/utils/recycleOCBuilds
+++ b/utils/recycleOCBuilds
@@ -1,0 +1,30 @@
+#!groovy
+
+properties(
+        [
+                buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '15', daysToKeepStr: '', numToKeepStr: '15')),
+                disableConcurrentBuilds(),
+                pipelineTriggers([cron('0 8 * * *')])
+        ]
+)
+
+def stepName = null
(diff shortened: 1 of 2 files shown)


utils/recycleOCBuilds
```diff
@@ -0,0 +1,30 @@
...
+            stepName = 'Remove Completed Builds'
+            stage(stepName) {
+                // Garbage collect all completed oc builds with -build in their name
+                sh 'oc delete pod $(oc get pods | grep \' Completed \' | grep \'\\-build\' | awk \'{print $1}\')'
+                currentBuild.result = 'SUCCESS'
+            }
+        } catch (Throwable err) {
+                        currentBuild.description = "Delete Failure"
+                        error "Find and deletion of completed oc builds failed!"
+                        currentBuild.result = 'FAILURE'
```
scoheb: Execution will never get here...`error` will rethrow an exception

johnbieren: Thanks, fixed

johnbieren:
[merge]

PR #658, opened 2018-06-13
srpm matcher fix

johnbieren:


src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -1409,7 +1409,7 @@ def repoFromRequest(String request, String prefix) {
         def gitMatcher = request =~ /git.+?\/([a-z0-9A-Z_\-\+\.]+?)(?:\.git|\?|#).*/
         def buildMatcher = request =~ /(?:koji-shadow|cli-build).+?\/([a-zA-Z0-9\-_\+\.]+)-.*/
         def pkgMatcher = request =~ /^([a-zA-Z0-9\-_\+\.]+$)/
-        def srpmMatcher = request =~ /.+?\/([a-zA-Z0-9\-_\+]+?)-[0-9a-zA-Z\.]+-.+.src.rpm/
+        def srpmMatcher = request =~ /.+?\/([a-zA-Z0-9\-_\+]+)-[0-9a-zA-Z\.]+-[0-9\.].+.src.rpm/
```
joejstuart: Can the release part contain letters? Just wondering if a-zA-Z should be included in that.

p3ck: Yes it can.  Good catch Joe

johnbieren: I think that already works because of the .+

I think every release contains letters, like 200.fc29.src.rpm, the fc part is matched here too. So I guess I am not sure what changes functionally as only name is of interest. Release always starts with numbers though to the best of my knowledge

johnbieren:
[upstream-test]

src/org/centos/pipeline/PipelineUtils.groovy
joejstuart: So that will work for every case except when the release contains just one digit, since the .+ matches 1 or more characters.

joejstuart: and that's only a concern if we have something like this. xfig-3.2.6a-8.src.rpm

johnbieren: I think the release should always have a fc or el in it, so I think that is fine. I am going to merge it and if it ends up breaking will make the change you recommended

johnbieren:
[merge]

PR #660, opened 2018-06-14
Parse more srpms

johnbieren:
java-1.8.0-openjdk-1.8.0.172-6.b11.fc27.src.rpm gave us grief. I think because there is a . in the name field. To the best of my knowledge, this will fix it


johnbieren:
[merge]

PR #661, opened 2018-06-15
Use srcpkg not pkg

johnbieren:
/pkg/ works with cockpit, so I thought it was solved, but not with vim. We need /srcpkg/ instead


johnbieren:
[merge]

PR #662, opened 2018-06-15
Save playbook run output

johnbieren:
Wanted to avoid having it be .log so it is not confused with test output


johnbieren:
[upstream-test]

johnbieren:
It was said we don't need to set python3 anymore, but my testing shows we do.

PR #667, opened 2018-06-20
Hard code last known good rawhide image

johnbieren:
This PR should be reverted once rawhide latest boots again


johnbieren:
[upstream-merge]

PR #668, opened 2018-06-20
Match _ in version

johnbieren:
cldr-emoji-annotation-33.0.0_2-1.fc29.src.rpm

This is correct fix, right?


johnbieren:
Noted. Thanks. [merge]

PR #677, opened 2018-07-04
singleshost-test: refresh dnf cache

thrix:
To make sure latest bits are pulled in when building.

Important usually for fast-moving standard-test-roles updates.


johnbieren:
[merge]

PR #681, opened 2018-07-18
Set pipefail before running test

johnbieren:
Now that we tee the output, we need pipefail, or else the playbook command will never fail, which breaks our logic for finding if things were unstable or not


joejstuart:
lgtm

johnbieren:
[merge] it failed because the stage job couldnt connect to irc, but the pipeline passed

PR #685, opened 2018-07-20
do not use hardcoded compose name

bgoncalv:


ignatenkobrain:
isn't there some kind of "latest" symlink?

bgoncalv:
I don't think we need "latest" symlink as only 1 image is expected to be on the directory. This is just an workaround until the proper job that validates the image and saves it is in place.

PR #694, opened 2018-07-23
Install python2-dnf when necessary

johnbieren:
I see some failures for python2-dnf not being present on the qcow2 image, which makes the dnf module not work

### DIFF
--- a/config/Dockerfiles/singlehost-test/rpm-verify.yml
+++ b/config/Dockerfiles/singlehost-test/rpm-verify.yml
@@ -1,6 +1,11 @@
 ---
 - hosts: localhost
   tasks:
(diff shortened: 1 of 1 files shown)


dirgim:
Hmm, does it fail in the playbook on the "Check which rpms are installed" task or somewhere else?

johnbieren:
Yes, check which rpms are installed.
msg": "`python2-dnf` is not installed, but it is requiredfor the Ansible dnf module.

dirgim:
Then I think only shell installation will work there, as package will also fail on that task.

johnbieren:
Good point, I'll fix it

johnbieren:
[merge]

PR #698, opened 2018-08-09
Recycle error pods too

johnbieren:
Recycles pods in error or evicted state that are older than 6 hours old. Pod life is in m when minutes, then h for hours, then d for days, so I delete if it contains d, and if it contains h I make sure the integer > 6.


johnbieren:
@bgoncalv fyi

johnbieren:
[merge]

PR #704, opened 2018-08-13
Change 6h limit to 1h

johnbieren:
6 hours is simply not good enough for some of our projects. They need pods removed more quickly


johnbieren:
[merge]

PR #705, opened 2018-08-13
Remove spaces in pod output

johnbieren:
Because the oc get pods has four things in it separated by spaces, so the if statement complains about too many arguments. Since we are just grepping for text anyways, we can just delete all the spaces in the output to avoid wrestling with more quoting and variable expansion.


johnbieren:
[merge] stage failure was due to missed message on datagrepper, not library compilation

PR #714, opened 2018-08-24
retry to update STR 5 times, if can't update do not fail test execution

bgoncalv:
If STR update fails, continue test execution with already installed STR version.


johnbieren:
[merge]

PR #716, opened 2018-08-27
Add clear template names function

johnbieren:
If you create podTemplate A then podTemplate B, podTemplate B will include all containers from A unless you call this before B's creation.


johnbieren:
Stage failures not due to this change

johnbieren:
[merge]

PR #717, opened 2018-08-28
Define resources for the singlehost-test container

psss:
Recently there were several jobs stuck in the pipeline in the test
execution step. The problem seems to be caused by insufficient memory.
Specifying minimum/maximum cpu/memory resources should fix that.


Andrei-Stepanov:
looks good, thank you!

johnbieren:
/me notes this is for the atomic pipeline, but it doesn't hurt to have. @psss would you like me to merge this?

psss:
Well, I submitted the PR to a wrong repo, the correct should be:
https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline/pull/113
But, agreed, this makes sense for Atomic pipeline as well. So +1 to merge it.

johnbieren:
[merge]

PR #718, opened 2018-08-28
Update executeInContainer with optional vars to be passed in.

p3ck:
This is actually from rob's PR 455.  It's needed when running
in parallel.  env is shared.


p3ck:
[merge]

PR #719, opened 2018-08-28
Use the name of the pod passed in..

p3ck:


p3ck:
seems there is a difference between podname and env.NODE_NAME.  It's unfortunate that the kuburnetes plugin adds an additional uuid to the end.

So this can't be merged. not yet anyway.

PR #720, opened 2018-08-29
Add ocVerifyPod and ocGetLogsFromPods

p3ck:
These methods assume you already called openshift.withcluster() and
openshift.withproject().  These should be used when doing parallel
jobs.

### DIFF
--- a/src/org/centos/pipeline/PipelineUtils.groovy
+++ b/src/org/centos/pipeline/PipelineUtils.groovy
@@ -758,6 +758,46 @@ def executeInContainer(String stageName,
 
 /**
  *
+ * @param nodeName podName we are going to verify.
+ * @return
+ */
+def ocVerifyPod(String nodeName) {
+    def describeStr = openshift.selector("pods", nodeName).describe()
+    out = describeStr.out.trim()
+
+    sh 'mkdir -p podInfo'
+
+    writeFile file: 'podInfo/node-pod-description-' + nodeName + '.txt',
+                text: out
+    archiveArtifacts 'podInfo/node-pod-description-' + nodeName + '.txt'
+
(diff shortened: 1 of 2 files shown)


p3ck:
[merge]

src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -756,6 +756,46 @@ def executeInContainer(String stageName,
...
 }
 
+/**
+ *
+ * @param nodeName podName we are going to verify.
+ * @return
+ */
+def ocVerifyPod(String nodeName) {
+    def describeStr = openshift.selector("pods", nodeName).describe()
+    out = describeStr.out.trim()
```
robnester-rh: I _think_ you need a ```def``` here?

src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -756,6 +756,46 @@ def executeInContainer(String stageName,
...
+
+    sh 'mkdir -p podInfo'
+
+    writeFile file: 'podInfo/node-pod-description-' + nodeName + '.txt',
```
robnester-rh: nit: it's more groovy-esque to do ```"podInfo/node-pod-description-${nodeName}.txt"```

src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -756,6 +756,46 @@ def executeInContainer(String stageName,
...
+                text: out
+    archiveArtifacts 'podInfo/node-pod-description-' + nodeName + '.txt'
```
robnester-rh: nit: as above.

PR #723, opened 2018-08-31
use the github api

joejstuart:
Instead of parsing the html, use the github api.

### DIFF
--- a/dev_setup/playbooks/roles/minishift/defaults/main.yml
+++ b/dev_setup/playbooks/roles/minishift/defaults/main.yml
@@ -4,6 +4,8 @@
 # default location for minishift
 minishift_dest_dir: "{{ ansible_env.HOME }}/minishift"
 
+minishift_version: v1.23.0
+
 # default minishift profile name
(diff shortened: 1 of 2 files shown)


scoheb:
Will you port this to contra-env-setup?

arilivigni:
@joejstuart you were going to add this back into the contra-lib correct?

joejstuart:
@scoheb yes, Is there a reason for it to exist here even?
@arilivigni I'm not following. The contra-lib stuff is just shared libraries right now.

arilivigni:
@joejstuart sorry I didn't look at was added.   I think I'm going to remove the dev_setup since folks get confused sometimes.   I saw your PR in the other repo will review

arilivigni:
@joejstuart speaking of libraries can we add library setup and use in https://githib.com/CentOS-PaaS-SIG/contra-env-setup?

joejstuart:
@arilivigni yes, good idea. I'll put a card in for it.

PR #676, opened 2018-07-03
Change singlehost test container to centos7

johnbieren:
Signed-off-by: Johnny Bieren <jbieren@redhat.com>

PR #728, opened 2018-09-06
singlehost-test: Move to Fedora base image

thrix:
We need to move back to Fedora, because of "Recommends:"
problem we see all over and they are not planning to backport
it to Centos:7


johnbieren:
https://github.com/CentOS-PaaS-SIG/ci-pipeline/pull/676

Might want to change the yum to dnf in the scripts as well, but it is just a nit pick. Assuming the python-dnf does not need to be versioned in the Dockerfile install, lgtm

PR #731, opened 2018-09-11
add repository with latest packages build in koji.

bgoncalv:
The repo has lower priority, this should avoid errors trying to install packages that are not in official repos.

Also enable updates-testing repo if it is not in Rawhide.

### DIFF
--- a/config/Dockerfiles/cloud-image-compose/virt-customize.sh
+++ b/config/Dockerfiles/cloud-image-compose/virt-customize.sh
@@ -69,7 +69,22 @@ enabled=1
 gpgcheck=0
 EOF
 
-virt-copy-in -a ${DOWNLOADED_IMAGE_LOCATION} ${CURRENTDIR}/testrepo/${package} ${CURRENTDIR}/test-${package}.repo /etc/yum.repos.d/
+koji_repo=${branch}
+if [ "${branch}" != "rawhide" ]; then
+    koji_repo="f${branch}-build"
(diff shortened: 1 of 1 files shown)


bgoncalv:
More info about latest fedora repos: https://fedoraproject.org/wiki/Repositories

thrix:
the problem here is, that the same packages will be always prioritized from the production repos. Only packages not found in the production repos will be installed from here. So I do not thing this makes sense.

What we need is what I tried to do on downstream:
Identify which koji builds we need to add to production repos, to be able to install the koji build. Then add those package with the same priority (or even higher priority - note means lower number priority in yum conf), so they are actually installed.

thrix:
Ok, I tested, and this works nicely also. No need to do magic scripting :) let's bring this in.

thrix:
@johnbieren this is good to go :) are you ok also?

thrix:
> the problem here is, that the same packages will be always prioritized from the production repos. Only packages not found in the production repos will be installed from here. So I do not think this makes sense.
> 
> What we need is what I tried to do on downstream:
> Identify which koji builds we need to add to production repos, to be able to install the koji build. Then add those package with the same priority (or even higher priority - note means lower number priority in yum conf), so they are actually installed.

scratch this, this is not true. the packages are chosen only from valid candidates really.

johnbieren:
@thrix yeah, looks fine to me. Feel free to merge

bgoncalv:
[upstream-merge]

PR #732, opened 2018-09-12
retry to send topics in case of failure

bgoncalv:
sometimes there could be some temporary issue when sending message to provider, the pipeline should retry to send the message.


bgoncalv:
[merge]

thrix:
yeah, retry

PR #736, opened 2018-09-17
catch exception when sending message

bgoncalv:
it looks like retry does not work if there is an exception, catch exception to see if problem is solved.

https://issues.jenkins-ci.org/browse/JENKINS-51454?jql=labels%20%3D%20declarative


bgoncalv:
[merge]

PR #741, opened 2018-09-25
exit with error if tests/inventory is not executable

bgoncalv:
STI only uses the inventory file as script so it has to be executable, so make sure pipeline does not execute the playbook using inventory with wrong permission.


Andrei-Stepanov:
ACK

bgoncalv:
[merge]

PR #743, opened 2018-10-01
Add str-container container

johnbieren:
This container is for running standard-test-roles against the container namespace instead of rpms. The major difference is that it needs to be able to run docker inside of it


config/Dockerfiles/str-container-test/container-test.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+set -e
```
joejstuart: Is the -x needed to print to console?

johnbieren:
[merge]

PR #751, opened 2018-10-11
Fix str-container-test container

johnbieren:
With Fedora 28, the pod is created and it fails instantly with "The agent has not been fully initialized yet." For some reason, I don't see this with Fedora 27. Also, with the Fedora 27 container, I have confirmed I don't need all the systemd hacking, /usr/sbin/init with the no VOLUME line seems to be sufficient.


johnbieren:
[merge] That sounds like it could be exactly it, but I don't know how to remedy that...

PR #752, opened 2018-10-15
nvr verify checks if system uses correct kernel if package is kernel …

bgoncalv:
…or kernel-rt

When testing kernel or kernel-rt packages, make sure VM uses correct kernel.

Also saves kernel being used to artifacts


bgoncalv:
[merge]

PR #754, opened 2018-10-30
allow insecure URL during trackMessage

bgoncalv:


bgoncalv:
@thrix this is not the only place in the Pipeline that curl runs with insecure, not sure why here we would use it as an option and in the other part not...

bgoncalv:
[merge]

PR #755, opened 2018-11-06
handle exception from sendCIMessage

bgoncalv:


thrix:
LGTM, would be nice if @johnbieren could check, that he thinks about it.

@johnbieren we sent the issue on IRC

bgoncalv:
[merge]

PR #756, opened 2018-11-14
Do not copy absent logs in executeInContainer()

Andrei-Stepanov:
Signed-off-by: Andrei Stepanov <astepano@redhat.com>

### DIFF
--- a/src/org/centos/pipeline/PipelineUtils.groovy
+++ b/src/org/centos/pipeline/PipelineUtils.groovy
@@ -772,7 +772,13 @@ def executeInContainer(String stageName,
     } catch (err) {
         throw err
     } finally {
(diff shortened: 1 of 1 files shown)


johnbieren:
@joejstuart This change seems fine, right? @Andrei-Stepanov it should be fine to merge despite the stage job results. A lot of pipelines use this function so I would like Joe to agree, but seems totally sane to me

Andrei-Stepanov:
@joejstuart hi!

yes, it just to remove message that is very confusing from builds logs:

```
15:37:20 + mv -vf logs brew-build/logs
15:37:20 mv: cannot stat 'logs': No such file or directory
```

johnbieren:
[merge]

PR #758, opened 2018-11-21
Fedora upstream will support PRs on tests namespace

bgoncalv:
For Fedora upstream pipeline to support PR on tests namespace these changes are required

Some trailing whitespaces got removed as well.

### DIFF
--- a/config/Dockerfiles/cloud-image-compose/virt-customize.sh
+++ b/config/Dockerfiles/cloud-image-compose/virt-customize.sh
@@ -22,6 +22,8 @@ virtlogd &
 
 chmod 666 /dev/kvm
 
+namespace=${namespace:-"rpms"}
+
 if [ $branch != "rawhide" ]; then
     branch=${branch:1}
 fi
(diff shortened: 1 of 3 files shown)


bgoncalv:
[upstream-test]

bgoncalv:
@johnbieren does this make sense? :)

johnbieren:
So if namespace = tests, the dir where the tests live is just different? That makes sense to me. Won't triggers have to change to listen to the additional namespace as well?

bgoncalv:
> So if namespace = tests, the dir where the tests live is just different? That makes sense to me. 
Yes, the idea if tests.yml will be defined directly on root of the repo, like https://src.fedoraproject.org/tests/shell/tree/master there there would be tests.yml there. @psss is it correct?

>Won't triggers have to change to listen to the additional namespace as well?
Yes, on `upstream-fedora-pipeline` we will need to include new trigger for it, also some other changes to the main Jenkinsfile to handle it, and I believe some changes related to the message sent to fedmsg. But I haven't had time to check it into details yet.

psss:
> Yes, the idea if tests.yml will be defined directly on root of the repo, like https://src.fedoraproject.org/tests/shell/tree/master there there would be tests.yml there. @psss is it correct?

Yes, exactly. The `tests.yml` file should be in the root of the test namespace directory so that tests then can be referenced from it directly.

bgoncalv:
[upstream-merge] [merge]

bgoncalv:
@johnbieren fyi, singlehost-test image didn't get promoted automatically, we needed to tag it manually.

johnbieren:
@bgoncalv yeah, the upstream-merge and merge trigger two separate merge jobs. One promotes cloud-image-compose, the other promotes singlehost-test (among others). No good answer is in place for PRs that affect both. I know you commented both merge commands (wisely), but it seems that whichever merge job goes first actually merges the PR and continues and the other does nothing I guess

PR #759, opened 2018-11-21
standard-test-roles does not require restraint-rhts any more, no need to install it

bgoncalv:


johnbieren:
Upstream Fedora Stage Pipeline Build failure seems unrelated. I'd like to let the Atomic one finish before merging

johnbieren:
[merge]

PR #764, opened 2019-01-29
make sure fedora rawhide has GPG key imported

bgoncalv:
attempt to solve issues like: https://jenkins-continuous-infra.apps.ci.centos.org/view/Fedora%20All%20Packages%20Pipeline/job/fedora-rawhide-build-pipeline/2165/artifact/package-tests/logs/FAIL_str_rust-cargo-test.log

### DIFF
--- a/config/Dockerfiles/cloud-image-compose/virt-customize.sh
+++ b/config/Dockerfiles/cloud-image-compose/virt-customize.sh
@@ -81,6 +81,8 @@ if [ "${branch}" != "rawhide" ]; then
         echo "failure enabling updates-testing repo"
         exit 1
     fi
+else
+    virt-customize -a ${DOWNLOADED_IMAGE_LOCATION} --run-command "source /etc/os-release; rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-\$VERSION_ID-fedora"
 fi
 # Add repo from latest packages built in koji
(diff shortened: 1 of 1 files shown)


bgoncalv:
It seems there is some issue with stage environment.

Anyway, it seems to work on prod: https://jenkins-continuous-infra.apps.ci.centos.org/view/Fedora%20All%20Packages%20Pipeline/job/fedora-rawhide-build-pipeline/2173/


@johnbieren @joejstuart what do you think? Should this be done at the image test for Rawhide instead, if so I'm not really sure how to do it :)

johnbieren:
It seems it may be able to go https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline/blob/master/vars/fedoraImageTest.groovy#L22 in the testCompose function, but I am not even sure where testCompose is defined. I am fine with you putting the change where you have it here.

johnbieren:
Stage is messed up because it is having trouble publishing on fedmsg stage. I'll check it out

johnbieren:
[upstream-test] [test]

johnbieren:
Seems to be fine. The atomic stage job failed the mockbuild it seems, but the all packages stage job finished unstable, which I think it typically does

bgoncalv:
[upstream-merge]

PR #765, opened 2019-01-31
Updates for jms-messaging-1.1.1

johnbieren:
We need jms-messaging-1.1.1 to send out nested arrays on fedmsg. These are the changes to resolve the syntax changes between 1.0.36 and  1.1.1

We can no longer have two topics for one trigger job, so some job(s) had to be split.

### DIFF
--- a/JenkinsfileTrigger
+++ b/JenkinsfileTrigger
@@ -1,3 +1,5 @@
+import groovy.json.JsonOutput
+
 properties(
         [
                 buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '50', daysToKeepStr: '', numToKeepStr: '100')),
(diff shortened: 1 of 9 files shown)


johnbieren:
Should be merged around the same time as 
https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline/pull/128
and
https://github.com/CentOS-PaaS-SIG/linchpin/pull/923
as the master needs to be upgraded to 1.1.1 and all three pertain to workflows on the same master.

johnbieren:
/me notes this passed stage on a custom master with a jms-messaging-1.1.2 snapshot plugin installed (there was another bug stopping us with 1.1.1). Once the new one is available to be installed, I can merge these and upgrade the Jenkins master.

johnbieren:
[merge]

PR #766, opened 2019-02-04
use --list-tags from ansible to check if specific tag is set

bgoncalv:


bgoncalv:
Use of ansible to check for tags should make easier to check for tags, form example if it is set like:

```tags: [classic, container, atomic]```

bgoncalv:
@johnbieren I think ansible is already installed on slave, isn't it?
https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/config/s2i/jenkins/slave/Dockerfile
https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline/blob/master/config/s2i/jenkins/slave/Dockerfile

johnbieren:
@bgoncalv Sorry, I wasn't suggesting that I am aware of anywhere currently that uses the function and is run from a slave without ansible. Just pointing out the technical consequence of the change. If using grep is not truly correct and it is only truly correct with ansible, I am fine with this change and having the function be dependent on ansible being present

bgoncalv:
[merge]

PR #768, opened 2019-02-04
revert changes to checkTests

bgoncalv:


bgoncalv:
revert the change until we have better way to check tags using ansible and STR

bgoncalv:
[merge]

johnbieren:
I am still trying to find out the stage job

PR #769, opened 2019-02-04
new version of STR split inventory as separate rpms

bgoncalv:


bgoncalv:
@johnbieren will you need qemu on str-container-test?

johnbieren:
@bgoncalv Not sure, I think you might know better than me. str-container-test is for running STR against the container namespace instead of the rpms namespace. Admittedly, I haven't looked at it much since implementing it as it seemed to just work and no one has said anything about it since.

johnbieren:
The stage jobs seem to be running fine now. I will merge this and make sure the stable tags are updated accordingly (I don't believe I put any automation in place for the str-container-test tags). [merge] fwiw the stage failures seem to be the tests for the upstream fedora pipeline failing (I see pretty much all of them failing in the stage job output) and the atomic one the canned ci message rpm isn't building, so perhaps I should update the canned message for the atomic pipeline. It is still tied to f27 though so I am not sure if we are getting too much value besides syntax checking out of that one anyways.

PR #771, opened 2019-02-05
avoid installing package that will conflict with another package provided by same koji build

bgoncalv:


bgoncalv:
it works: https://jenkins-continuous-infra.apps.ci.centos.org/job/fedora-rawhide-stage-pr-pipeline/284 [upstream-merge]

PR #773, opened 2019-02-07
use ansible + str to check of tag on playbooks

bgoncalv:
If standard-test-roles is available use it to check for tag in the playbooks

### DIFF
--- a/upstream-fedora-pipeline/AllPackagePipelineStageTrigger
+++ b/upstream-fedora-pipeline/AllPackagePipelineStageTrigger
@@ -101,6 +101,7 @@ pipeline {
                     changeset "config/Dockerfiles/rpmbuild/**"
                     changeset "config/Dockerfiles/cloud-image-compose/**"
                     changeset "config/Dockerfiles/singlehost-test/**"
+                    changeset "config/s2i/jenkins/slave/**"
                     changeset "upstream-fedora-pipeline/**"
                 }
             }
@@ -117,6 +118,18 @@ pipeline {
             steps {
                 script {
                     pipelineUtils.setupContainerTemplates(openshiftProject)
+                }
+            }
+        }
+        stage("jenkins-continuous-infra-slave image build") {
+            when {
+                // Only build if we have related files in changeset
+                changeset "config/s2i/jenkins/slave/**"
+                environment name: 'relevant', value: 'true'
+            }
+            steps {
+                script {
+                    tagMap['jenkins-continuous-infra-slave'] = pipelineUtils.buildImage(openshiftProject, "jenkins-continuous-infra-slave")
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


johnbieren:
@bgoncalv I'd recommend you add the path for the slave directory around here
https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/upstream-fedora-pipeline/AllPackagePipelineStageTrigger#L104
Because of the state of the f27 atomic stage pipeline, it doesn't get far enough to try checkTests and the upstream-fedora-pipeline is dropping this change as not relevant because of ^^

bgoncalv:
thanks, I've updated it

upstream-fedora-pipeline/AllPackagePipelineStageTrigger
```diff
@@ -101,6 +101,7 @@ pipeline {
                     changeset "config/Dockerfiles/rpmbuild/**"
                     changeset "config/Dockerfiles/cloud-image-compose/**"
                     changeset "config/Dockerfiles/singlehost-test/**"
+                    changeset "config/s2i/jenkins/slave/**"
```
johnbieren: Sorry, I forgot until now, but something like this
https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/upstream-fedora-pipeline/AllPackagePipelineStageTrigger#L123..L134
for the slave container will also need to be added (so the tagMap is updated with the new tag and that is passed to build job:)

johnbieren:
https://jenkins-continuous-infra.apps.ci.centos.org/job/test-for-bruno/1/console
```
19:10:16 [test-for-bruno] Running shell script
19:10:17 + grep -e 'TASK TAGS: \[.*\<classic\>.*\]'
19:10:17 + ansible-playbook --list-tags standard-test-roles/tests/tests.yml
19:10:17  [WARNING]: provided hosts list is empty, only localhost is available. Note
19:10:17 that the implicit localhost does not match 'all'
19:10:18       TASK TAGS: [classic]
[Pipeline] echo
19:10:18 Starting package-tests
```

Is this what is expected?

bgoncalv:
> https://jenkins-continuous-infra.apps.ci.centos.org/job/test-for-bruno/1/console
> 
> ```
> 19:10:16 [test-for-bruno] Running shell script
> 19:10:17 + grep -e 'TASK TAGS: \[.*\<classic\>.*\]'
> 19:10:17 + ansible-playbook --list-tags standard-test-roles/tests/tests.yml
> 19:10:17  [WARNING]: provided hosts list is empty, only localhost is available. Note
> 19:10:17 that the implicit localhost does not match 'all'
> 19:10:18       TASK TAGS: [classic]
> [Pipeline] echo
> 19:10:18 Starting package-tests
> ```
> 
> Is this what is expected?

Yes, it is some warning from ansible. If you think this can be confusing I can redirect the output...

johnbieren:
Tbh, I interact with these pipelines very little nowadays (just for messaging and to help you guys in the rare times you need it!) so whatever output you think is best is fine by me. 

If that stage job looks good I am fine with you merging this. I have an idea for how to have the shared lib changes that affect the upstream-fedora-pipeline be tested properly... on my todo list to try out

thrix:
@bgoncalv  I believe we can live with another warning

bgoncalv:
[merge]

PR #777, opened 2019-02-15
Back to jms-1.1.2 format

johnbieren:
Might need to add a commit for jms messaging version as I am waiting on a recently merged PR in a release

Had to remove topic from the messages as it prevents the messages from being sent on the bus


johnbieren:
[merge]

PR #791, opened 2019-03-08
Workaround socket timeout losing --wait

johnbieren:
We keep getting java socket timeouts when oc start-build begins, so the --wait is lost and the build proceeds with an unfinished image, which fails. This seems to work around it.


johnbieren:
@bgoncalv

johnbieren:
[merge]

PR #793, opened 2019-03-14
Allow for custom stable tags to be used

p3ck:
This is helpful if the project has multiple development branches
at the same time.


johnbieren:
[merge]

PR #797, opened 2019-03-26
Check if any rpms satisfy conflict capability instead

johnbieren:
We remove awk from the first repoquery to save the entire capability. Then, we perform another repoquery to see if any of the rpms in the repo meet that capability, which would mean there is a conflict. Seems to work in my local testing.

### DIFF
--- a/config/Dockerfiles/cloud-image-compose/virt-customize.sh
+++ b/config/Dockerfiles/cloud-image-compose/virt-customize.sh
@@ -109,7 +109,8 @@ virt-copy-in -a ${DOWNLOADED_IMAGE_LOCATION} ${virt_copy_files} /etc/yum.repos.d
 if [ "${namespace}" != "tests" ]; then
     for pkg in $(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --all --qf="%{ARCH}:%{NAME}" | sed -e "/^src:/d;/-debug\(info\|source\)\$/d;s/.\+://" | sort -u) ; do
         # check if this package conflicts with any other package from RPM_LIST
-        conflict=$(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --conflict $pkg | awk '{print$1}')
+        conflict_capability=$(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --conflict $pkg)
+        conflict=$(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --whatprovides "$conflict_capability" | awk '{print$1}')
         found_conflict=0
         if [ ! -z "${conflict}" ] && [ ! -z "${RPM_LIST}" ]; then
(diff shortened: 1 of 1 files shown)


johnbieren:
@bgoncalv what do you think?

bgoncalv:
nice! I tested with the ignition build and both packages got installed:

https://jenkins-continuous-infra.apps.ci.centos.org/job/fedora-rawhide-stage-pr-pipeline/342/

also tested with curl, where curl-minimal doesn't get installed as expected:

https://jenkins-continuous-infra.apps.ci.centos.org/job/fedora-rawhide-stage-pr-pipeline/343/

config/Dockerfiles/cloud-image-compose/virt-customize.sh
```diff
@@ -109,7 +109,8 @@ virt-copy-in -a ${DOWNLOADED_IMAGE_LOCATION} ${virt_copy_files} /etc/yum.repos.d
 if [ "${namespace}" != "tests" ]; then
     for pkg in $(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --all --qf="%{ARCH}:%{NAME}" | sed -e "/^src:/d;/-debug\(info\|source\)\$/d;s/.\+://" | sort -u) ; do
         # check if this package conflicts with any other package from RPM_LIST
-        conflict=$(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --conflict $pkg | awk '{print$1}')
+        conflict_capability=$(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --conflict $pkg)
+        conflict=$(repoquery -q --disablerepo=\* --enablerepo=${package} --repofrompath=${package},${rpm_repo} --whatprovides "$conflict_capability" | awk '{print$1}')
```
mike-nguyen: The result of this query is the NEVRA of package that matches so `awk '{print$1}` also just prints out the NEVRA.  I think the the comparison on line 117 is just comparing the package name so we probably want to add -`-qf "%{NAME}"` to this query.

johnbieren:
@bgoncalv so should I merge this? The stage job failed in the verify-rpm step... not sure what the status of the canned message is currently, but that wouldn't be related to this change

bgoncalv:
> @bgoncalv so should I merge this? The stage job failed in the verify-rpm step... not sure what the status of the canned message is currently, but that wouldn't be related to this change

yes, you can merge. It fails on nvr verify, because the canned message builds some old standard-test-roles version and we end up installing a new version from STR module... To fix that I think we would need to use a different package for the canned message.

johnbieren:
Ah, right, I remember that now. [upstream-merge]

config/Dockerfiles/cloud-image-compose/virt-customize.sh
johnbieren: For some reason this did not show up for me until now... Looking. I will open another PR if needed

johnbieren: Yes, I agree. Opening another PR

PR #799, opened 2019-03-28
Move cat of file in obtainLock function

johnbieren:
@bgoncalv 

Makes the super unlucky timing situation even less likely


johnbieren:
[merge]

PR #801, opened 2019-04-03
virt-customize uses dnf to install packages

bgoncalv:
Fix for https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline/issues/140


bgoncalv:
[upstream-merge]

PR #802, opened 2019-04-09
sendMessageWithAudit: make optional to check message in datagrepper

bgoncalv:


bgoncalv:
@johnbieren could you check this?

bgoncalv:
The pipeline failed, but due to unrelated issue.

johnbieren:
I don't like this change. If you don't want to track the message, sendMessage should be used instead. The only difference between sendMessage and sendMessageWithAudit is tracking the message

bgoncalv:
hmm, okay. I can use sendMessage.

the sendMessageWithAudit also creates the audit file.

Feel free to close the PR

PR #804, opened 2019-04-16
Get more debug info from create-containers.sh

14rcole:
This is a change that was made a while ago while herlo was working on linchpin and a PR was never opened.  If this gets merged we can switch to this repo instead of his fork

### DIFF
--- a/src/org/centos/pipeline/PipelineUtils.groovy
+++ b/src/org/centos/pipeline/PipelineUtils.groovy
@@ -955,7 +955,7 @@ def setupContainerTemplates(String openshiftProject) {
     openshift.withCluster() {
         openshift.withProject(openshiftProject) {
             dir('config/s2i') {
-                sh './create-containers.sh'
+                sh 'bash -x ./create-containers.sh'
             }
         }
     }


14rcole:
[test]

14rcole:
@robnester-rh the build seems to be failing at the `singlehost-test image build step`.  Looks like some sort of a network timeout.  Is there a way to get more info from the thread building the image?

johnbieren:
It is failing in the stage pipeline, not an openshift build image step. It is a known issue, not related to this PR. Nothing past the point where it fails would be affected by this PR anyway

src/org/centos/pipeline/PipelineUtils.groovy
```diff
@@ -955,7 +955,7 @@ def setupContainerTemplates(String openshiftProject) {
     openshift.withCluster() {
         openshift.withProject(openshiftProject) {
             dir('config/s2i') {
-                sh './create-containers.sh'
+                sh 'bash -x ./create-containers.sh'
```
johnbieren: We can merge this, but I don't really see the point. Functionally, this is not going to change anything, so there is no way that this is currently blocking LP release from working and that merging it would fix it. It will just give us more output in the jenkins logs. It does not matter much to me if we merge it

PR #805, opened 2019-04-25
Don't fail on oc delete pod either

johnbieren:
This fails if ```oc get pods | grep ``` returns nothing as well. Since they all will have OR true attached to them, I removed the error catching too.

@bgoncalv


bgoncalv:
lgtm

PR #810, opened 2019-05-09
retry if can't read lock file

bgoncalv:
if for some reason the file existed, but could not be read, just retry...


bgoncalv:
[merge]

PR #821, opened 2019-08-09
force sync artifacts in the end of test execution

bgoncalv:
Make sure artifacts are synced after tests are executed,
even if test playbook didn't do it or got aborted


bgoncalv:
[merge]

PR #822, opened 2019-08-15
Add description to the shell scripts

Turivniy:


Turivniy:
It will update steps with the description in the Blue Ocean stages.
All of the "Shall Script" steps will be updated with the description.

johnbieren:
@bgoncalv wdyt? It lgtm

bgoncalv:
@johnbieren it lgtm, I just don't know how many of this function we are actually using :)
Like some moved to contra-lib other to upstream-first repo....

https://github.com/openshift/contra-lib/blob/master/src/org/centos/contra/pipeline/Utils.groovy

https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline/blob/master/src/org/centos/pipeline/PackagePipelineUtils.groovy

johnbieren:
Agreed, but might as well [merge] I suppose

PR #829, opened 2020-04-08
Proposed fix for Fedora CI issue 104

jimbair:
https://pagure.io/fedora-ci/general/issue/104

Our cluster is openshift 3.6 and requires the 3.6 cli per pagure fedora-ci issue 104

Also, a quick glance shows other images related to jenkins are also still on 3.6, which is expected given our cluster:
https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/config/s2i/jenkins/slave/Dockerfile#L1


bgoncalv:
[merge]

PR #830, opened 2020-04-14
Lock plugins to current versions while fixing fedora-ci issue 104

jimbair:
This is a follow-up to https://github.com/CentOS-PaaS-SIG/ci-pipeline/pull/829

While trying to re-deploy in openshift, it started upgrading our plugins. While we do want to do this, we don't want to both upgrade every plugin AND try to fix our CLI tools. =) One problem at a time. Why this didn't trigger until now but we managed to upgrade our CLI I cannot say.


jimbair:
[merge]

